Issue #31, opened 2025-07-30
[BUG] convex hits does not work with a perfectly vertical (or horizontal) boundary


BaronAWC:
**Describe the bug**
when using convex hits between 2 clips, if a boundary of a clip is perfectly vertical, it does not detect an intersection.

**To Reproduce**
see wick file

**Expected behavior**
intersections are displayed on vertical boundaries

**Screenshots**
if it detected an intersection, the red square would be positioned there.
<img width="714" height="326" alt="Image" src="https://github.com/user-attachments/assets/2a37571f-71e3-4683-9147-6710792e4bbb" />

**Desktop (please complete the following information):**
macOS chromium, 1.19.3 (not 1.19.4 but i cant bother checking right now)

**Do you have a suggested solution to this issue? (ex. has another program fixed this bug a certain way? Are you familiar with where in the code base someone would need to fix this issue?)**

**Additional context**
wick file is here: https://files.catbox.moe/qj281c.wick

PR #38, opened 2025-08-16
Convex hits fix

Alani25:
Doesn't 100% fix the convex collision but does improve it. I realized there might be issues with the actual algorithm used that we might want to look through to fix the issue with collision between flat edges of shapes (convex collision). 

Original hits convex code written by nick, using this algo:
https://www.bowdoin.edu/~ltoma/teaching/cs3250-CompGeom/spring17/Lectures/cg-convexintersection.pdf

See commit https://github.com/Candlestickers/Candlestick/commit/a588e13f0a50d0843dd80eba6087876f82055256
See issue https://github.com/Candlestickers/Candlestick/issues/31

Test file (created by BaronAWC, edited slightly by me): [HitsBug.zip](https://github.com/user-attachments/files/21815920/HitsBug.zip)

### DIFF
--- a/engine/dist/emptyproject.html
+++ b/engine/dist/emptyproject.html
@@ -14,7 +14,7 @@
         <!-- The Wick Engine build is injected here during the engine build step (see gulpfile.js) -->
         <script>
             /*Wick Engine https://github.com/Wicklets/wick-engine*/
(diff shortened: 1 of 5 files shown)


Alani25:
NOTE: Do not delete the convexHitsFix branch, we'll likely need to come back to it in the future

PR #42, opened 2025-08-21
41 Adding WickSound

jovannyrodz:


BaronAWC:
possible additions

1. get rid of `currently()` and overload `seek()` to perform the same thing as `currently()` if nothing is passed in.
2. add other methods like altering the speed/rate of an audio file. basically, just turning something like `project.mainSong.asset._howl.rate(1)` into a function.
    - an alternative is to make a shortcut for `asset._howl`, and just consider it a way to access more advanced howler functions, or just other ones like `rate()` or `fade()` that we don't currently have. so if people really need the functionality, they can just read the howler documentation.
3. better looping: the audio should be able to loop between any 2 points, not just the start and end.

i made a bit of a crappy implementation of this in a past project (which i don't think we should use - see why later - but it's here for reference).

this uses the end and length of the loop purely because it was more convenient for me at the time of making this modification. however, we can discuss whether it's better to do something like the loop start and end instead.

```
window.WickSound = class {
	constructor(assetName, loopEnd, loopTime) {
		this.asset = project.project.getAssetByName(assetName);
		this._pan    = 0;
		this._volume = 1;
		this._seek   = 0;
		this._loop   = false;
		this._muteVolume = 1;
		this.loopEnd = loopEnd; // where the loop ends
 		this.loopTime = loopTime; // seconds of looping
	}

        ...

        checkLoop() {
	    if(this.currently() > this.loopEnd) {
	        this.seek(this.currently() - this.loopTime);
	    }
	}

        ...
```

**the `checkLoop()` function was thrown into an update script for my game, so that it would check every frame and loop properly the moment it was supposed to. though, i have no idea if howler has a better way of doing this since i was in a rush. in any case, i don't think this is the best way to implement this, and if there's any way to not force the coder to put this in an update script, that would be ideal.**

jovannyrodz:
> possible additions
> 
> 1. add documentation in the "Sound" tab of the script editor. probably good to do before release. we can leave playSound and stopAllSounds since they are useful, but we should somehow explain that WickSound is good for more advanced sound control.
> 2. get rid of `currently()` and overload `seek()` to perform the same thing as `currently()` if nothing is passed in.
> 3. add other methods like altering the speed/rate of an audio file. basically, just turning something like `project.mainSong.asset._howl.rate(1)` into a function.
>    
>    * an alternative is to make a shortcut for `asset._howl`, and just consider it a way to access more advanced howler functions, or just other ones like `rate()` or `fade()` that we don't currently have. so if people really need the functionality, they can just read the howler documentation.
> 4. better looping: the audio should be able to loop between any 2 points, not just the start and end.

[1] That would be a separate issue for the UI.  
[2] currently() method is fine, I have seen in other sound.music APIs.  We can definitely overload seek without parameters as well, so the user can use both as they preferred.
[3] Since this is a wrapper, I just going to provide a way to get the Howler object, so people can use Howler if they want.
[4] I get the point of your Advance Loop, I'll be implementing that as a separate method.

src/Editor/EditorCore.jsx
```diff
@@ -474,7 +474,7 @@ class EditorCore extends Component {
    * @param {string|number} newValue  New value of the attribute to update.
    */
   setSelectionAttribute = (attribute, newValue) => {
-    this.project.selection[attribute] = newValue;
+    try{this.project.selection[attribute] = newValue;}catch(err){}
```
Alani25, line 477: My only comment is you can probably remove this part from EditorCore.jsx (the try catch statement) since we found the bug.

Alani25 (approved):
Your WickSound api makes working with audio thousand times easier in wick, and I love that when no input is given the functions return a value instead.
And if we plan to use this example file with the release, just make sure to credit the original song if not already.

Also I love the rate variable! Slowing down the sound to a rate of 0.4 really switched it from a happy song to something more sad, which goes to prove how effective this setting is.

Great job!

engine/dist/emptyproject.html
```diff
@@ -55658,6 +55658,176 @@
...
+  } // Method for rate control, rate = 1 is the default sound rate.
+  // rate can go from 0.1 to 4
+
+
+  rate(rate) {
+    if (rate === null || rate === undefined) {
+      return this.asset._howl.rate();
+    }
+
+    if (rate <= 0.1) rate = 0.1;else if (rate > 4) rate = 4;
```
BaronAWC, line 55745: even though yes, there isn't really a reason for the rate to go so low or high, i think it's kinda bad to set arbitrary caps like this. tape-stop effects (when the rate and pitch of a song transitions from 1 to 0) will be easier to code, and if someone wants to make something feel like a mega-timelapse with mega-fast and high-pitched audio, they may want to go higher than 4.

the only restriction i would add is that the rate should have a minimum value of 0, and only because howler doesn't let audio play in reverse.

engine/dist/emptyproject.html
```diff
@@ -55658,6 +55658,176 @@
...
+    this._pan = pan;
+    if (this._pan < -1) this._pan = -1;else if (this._pan > 1) this._pan = 1;
+
+    this.asset._howl.stereo(this._pan);
+  } // seek sound in seconds from 0 to sound duration
+
+
+  seek(seconds) {
+    if (seconds === null || seconds === undefined) {
+      return this.asset._howl.seek();
```
BaronAWC, line 55765: you could just call this.currently() instead

engine/dist/emptyproject.html
```diff
@@ -55658,6 +55658,176 @@
...
+    }
+  } // required only to update the loop when using custom loop range
+
+
+  updateLoop() {
+    if (this._loop && this._hasCustomLoopRange) {
+      if (this.isPlaying()) {
+        let curr = this.currently();
+
+        if (curr < this._loopRange.low || curr > this._loopRange.high - 0.01) {
```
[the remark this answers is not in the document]
BaronAWC: also, for the 2nd half of the if statement, should we be doing `curr >= this._loopRange.high` instead of doing the `- 0.01` thing?

engine/dist/emptyproject.html
jovannyrodz: Noted.  This is just for people that want to use our API, something greater than 4 is not useful or lower than 0.1.  People can always get the Howler instance from the new method and do more crazy things.  No change needed.

engine/dist/emptyproject.html
```diff
@@ -55658,6 +55658,176 @@
...
+    this._loopRange = {
+      low: 0,
+      high: this.duration()
+    };
+    this._hasCustomLoopRange = false;
+    this._muteVolume = 1;
+    this._loopTimer = null;
+  }
+
+  setLoopRange(loopRange) {
```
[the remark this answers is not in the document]
jovannyrodz: Yup, I like the song.setLoopRange(10, 20) API.  I will implement the change.

engine/dist/emptyproject.html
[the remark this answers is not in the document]
jovannyrodz: If generate better results, I'll remove it.

jovannyrodz:
> Your WickSound api makes working with audio thousand times easier in wick, and I love that when no input is given the functions return a value instead. And if we plan to use this example file with the release, just make sure to credit the original song if not already.
> 
> Also I love the rate variable! Slowing down the sound to a rate of 0.4 really switched it from a happy song to something more sad, which goes to prove how effective this setting is.
> 
> Great job!

Song credits added into the example.

PR #45, opened 2025-08-23
Theme rebranding

Alani25:
New theme colors added to the site (mainly UI stuff)


CNAME
BaronAWC, line 1: are we still putting "test" at the front?

package.json
BaronAWC, line 1: should we still have test.candlestickers.github.io here

BaronAWC (commented):
there are a few instances of test.candlestickers.github.io which i dont think should be there? otherwise this should be good

package.json
Alani25: Nope we shouldn't, that was based off the old way wick was set up ([test.wickeditor.com](https://test.wickeditor.com/)). Now we go with [candlestickers.github.io/Candlestick/test/](https://candlestickers.github.io/Candlestick/test/). It shouldn't be doing anything, I'll still make sure to ignore this part while merging.

BaronAWC (approved):
should be all set

jovannyrodz:
[1]. Ok, Can we have at the browser tab "WE Candlestick 1.0.0" or "WickEditor C 1.0.0"  probably the first one.
[2]  Same with the ghost info button at the upper left.
[3]. The alignment behavior for using align right or align bottom twice or more...

Everything else seems to be working fine and looks good.  Thank you for all hard the work. (30 commits and 40+ file changes)

PR #49, opened 2025-08-27
set $blockScrolling to Infinity instead of true

BaronAWC:


Alani25 (commented):
This is indeed revolutionary

Alani25 (approved):
This is indeed revolutionary.

Issue #52, opened 2025-08-28
Sub Clip Collision and Quadratree


jovannyrodz:
Sub Clip Collision and Quadratree

PR #54, opened 2025-08-28
Review for Subclip hits #52

jovannyrodz:
#52 
SubClip hits from 1.19.4

Branch name is wrong from content is ok.


engine/src/base/Project.js
```diff
@@ -288,18 +352,13 @@ Wick.Project = class extends Wick.Base {
 
     set hitTestOptions(options) {
         if (options) {
-            if (options.mode === 'CIRCLE' || options.mode === 'RECTANGLE' || options.mode === 'CONVEX') {
```
BaronAWC, line 291: the CIRCLE hitTestOptions type does not work properly in this version. and since CIRCLE is removed here, i wonder if this is where the issue is.

here's the demo that luxapodular made a while back. i had to rename it to a zip file so github would let me upload it, but change the file back to .wick.
[NewHitsDemo.zip](https://github.com/user-attachments/files/22034701/NewHitsDemo.zip)

this bug exists in 1.19.4 and this branch, but NOT 1.19.3 or the deployed version of candlestick.

BaronAWC (changes requested):

engine/src/base/Project.js
BaronAWC: ok looking at the diffs in this branch and in regular wick, it's definitely more than just this line. for some reason the CIRCLE type was removed in a lot of places.

jovannyrodz: I saw it yesterday, I have never work with the hitTestOptions... it looks like they still have the circle to circle and retangle to rectangle collision detection.

PR #57, opened 2025-08-30
Review 56 clip api for layer depths

jovannyrodz:
Adding dynamic ability to clips to move depths within a layer
Adding dynamic ability to clips to change from a layer into another.


Alani25 (changes requested):
Tested both in preview, preview window, and HTML/ ZIP exports, and code works as expected. 
I found the swapDepth function interesting in particular, it looks pretty useful.
There is one thing I'd like you to do, and that is to add these new functions with a short description under the code reference just so we have it documented somewhere the user can find and use easily.

image

Also while you're at it, either in this branch or another, add your sound code as well.
I believe the code for the reference is inside `src` >> `editor` >> `scriptInfo.js` and hopefully easy to add.

That aside all looks good 
Thanks for your hard work :)

jovannyrodz:
> pumpkinhead's name is misspelled (missing a p), and i think the sendForward() and sendBackward() were also made by him. other than this and what hamzah said, this should be all set.

Sure,  I'll fix his name, but I added my versions of sendForward() and sendBackward() functions.   I'll not put my name either.

jovannyrodz:
> Tested both in preview, preview window, and HTML/ ZIP exports, and code works as expected. I found the swapDepth function interesting in particular, it looks pretty useful. There is one thing I'd like you to do, and that is to add these new functions with a short description under the code reference just so we have it documented somewhere the user can find and use easily.
> 
> image
> Also while you're at it, either in this branch or another, add your sound code as well. I believe the code for the reference is inside `src` >> `editor` >> `scriptInfo.js` and hopefully easy to add.
> 
> That aside all looks good Thanks for your hard work :)

Ok, I just added two more methods and updated the scriptInfo.

engine/src/base/Clip.js
```diff
@@ -1334,6 +1334,154 @@ let avgIntersection = {
...
+    let index = this.depth;
+    
+    siblings.splice(index, 1);
+    siblings.push(this);
+  }
+
+  /**
+   * depth - Send clip to Absolute Front: top layer, top depth
+   */
+  sentToAbsoluteFront() {
```
BaronAWC, line 1404: should be send, not sent

src/Editor/scriptInfo.js
```diff
@@ -244,23 +244,66 @@ class ScriptInfoInterface extends Object {
...
+                    description: 'Moves this object to the bottom layer and to the back of that layer.'
+                  },
+                  {
+                    name: 'sendToFront',
+                    snippet: 'this.sendToFront()',
+                    description: 'Moves this object in front of all siblings in its current layer.'
+                  },
+                  {
+                    name: 'sentToAbsoluteFront',
+                    snippet: 'this.sentToAbsoluteFront()',
```
BaronAWC, line 269: same thing here, use send instead of sent

BaronAWC (changes requested):
all good other than renaming

Alani25 (approved):
Love the two new functions you added! 
Also thanks for adding them to the reference code, there's a lotta new functions they're hard to remember so that helps a lot. The detailed info you added under each makes them easy to understand as well for beginners.

Asides from Baron's notes and quick fixes I don't have much to add.
Great stuff :)

engine/src/base/Clip.js
jovannyrodz: Dang, LOL That was GPT5.

jovannyrodz: Nice catch

jovannyrodz:
> Love the two new functions you added! Also thanks for adding them to the reference code, there's a lotta new functions they're hard to remember so that helps a lot. The detailed info you added under each makes them easy to understand as well for beginners.
> 
> Asides from Baron's notes and quick fixes I don't have much to add. Great stuff :)

That part was done by gpt5 : )

PR #66, opened 2025-09-18
Mobile nav — Touch Gestures

Alani25:
This includes:
- Two finger zoom
- Two finger tap

Has been programmed to work with all the different tools (cursor to bucket) and tested with each. 

Can only be tested on mobile devices. 
When running `npm start` make sure to go to the network development server by visiting the given address on your touch device (ipad, phone) that are connected to the same network. 
image
And yes I did just leak my ip address— 
<sup>jk it's a private ip address on a private network, good luck using that to locate me.</sup>

I'll leave work for 2 finger tap undo for the next version.

### DIFF
--- a/engine/src/tools/Brush.js
+++ b/engine/src/tools/Brush.js
@@ -78,6 +78,23 @@ Wick.Tools.Brush = class extends Wick.Tool {
         return true;
     }
 
+    cancel () {
+        if (!this._isInProgress) return;
+        
+        this._suppressCommit = true;
+        this._gestureInterrupted = true;
+        
+        // Clear any pending work (potrace timeout)
+        if (this._croquisStartTimeout) {
+            clearTimeout(this._croquisStartTimeout);
+            this._croquisStartTimeout = null;
+        }
+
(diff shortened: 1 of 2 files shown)


Alani25:
Added in 2 finger click to undo, fixed previous issues, asides from zooming in isn't centered. I tried to improve it by combining panning with zooming when two fingers are "zooming in" to make it appear were zooming towards center of the two cursors, but had to limit it since when you reached max zoom it kept panning. 
Another annoying thing I realized is the way I (or anyone else) casually zooms in. You might move two fingers, or keep one still and zoom the other or what not. 
Max zoom limitation differed based on device. The improvement is still there in this version but I don't think it's major enough to be noticed or for us to worry about until later future versions. There are better things to worry about in terms of mobile development, like getting those file upload and export to work (which I'll start in later branch after we're sure this one is safe to release to development)

BaronAWC:
i'm still getting it to jump around if i'm pinching with the zoom tool, though it's definitely less frequent. still trying to figure out the exact cause.

PR #67, opened 2025-09-18
V1 release prep

Alani25:
Removed instances of Wick mascot and added in discord server button in place of previous "Support Us" button
image

image


src/Editor/_wickbrand.scss
BaronAWC, line 1: could rename patreon-orange and patreon-text to discord and discord-text

BaronAWC (approved):

src/Editor/_wickbrand.scss
Alani25: I decided to keep most of the old "support us" in there the same way it had been in case we decide to add in our own patreon and support section later on. Yeah I understand the naming doesn't follow the "standard," I'll make the fix accordingly

PR #87, opened 2025-10-23
Layer opacity

StickmanRed:
This includes inspector values and works with onion skinning. Tested locally. I didn't add the engine builds this time


engine/src/base/Layer.js
```diff
@@ -32,6 +32,7 @@ Wick.Layer = class extends Wick.Base {
 
         this.locked = args.locked === undefined ? false : args.locked;
         this.hidden = args.hidden === undefined ? false : args.hidden;
+        this.opacity = args.opacity;
```
BaronAWC: idk how important it is to add the undefined check, but you could just add it to be super safe

`this.opacity = args.opacity === undefined ? 1 : args.opacity;`

PR #91, opened 2025-11-04
Hamzah appdev

Alani25:
Using tauri to create an app version (windows & mac) of Candlestick

Comparison between the app version of Wick and our app version:

| | Candlestick | Wick |
| ---- | ----- | ---- |
| **Tool** | Tauri | Electron |
| **Size** | ~60 MB | ~600 MB |
| **Load Speed** | <5s | <12s |

Bugs in Wick app that have been fixed in Candlestick:
- Slight "spikes" in brush path
- Preview window not working
- Built-in assets not working (audio assets still need work in CS)
- No inspector tools

I personally think the major trade off is the app size difference and speed in performance as a result. That, and also with Electron you need to use SVGs for our UI to load (that explains all the SVGs). On top of that, of course, comes all the other fixes we made that were specific to the app version, without breaking the web version in the process.

There's more to be done so we'll keep this branch after merging with development, but I believe there's enough to at least get some alpha testing from other people (also there's a lotta demand and the code is already public so no point in hiding it from people).


Alani25:
> i think it would be good for the editor to fill the screen by default, or at least for the window to remember its size between sessions. not too big of a deal though, you can decide if that's worth your time right now.

Updated code to open the app in full screen by default

> though one thing we probably should add is a minimum window size. cuz it can get... small. so small that the only thing left is the traffic lights in the title bar.

I wouldn't worry about this — since the window isn't at minimum scale by default and if the user resizes it that small... well then that's more so on the user. Either way, I still set minimum window width and height to 320 and 200 px, respectively (gave it a low value bc I don't expect the user to go that low, but if they want to... they can).

StickmanRed (approved):
Submitting so the pr gets merged sooner, I'll test the app as soon as I can (the setup is still loading)

PR #101, opened 2026-01-01
fix: Editor crash using gotoFrame(NaN)

bogoddr:
Adds an `isNaN` check to all timeline methods that use `gotoFrame`.

I ran into this issue while doing some sketchy math to dynamically change the framerate of a clip. This is kind of an edge case, but it's better to throw a script error than have the editor fully crash.

**Test**

Running this project should not crash the editor. The project should stop immediately after running and show the error in the script editor.

[gotoframe NaN crash test1-1-2026_13-51-47.zip](https://github.com/user-attachments/files/24402519/gotoframe.NaN.crash.test1-1-2026_13-51-47.zip)

### DIFF
--- a/engine/dist/emptyproject.html
+++ b/engine/dist/emptyproject.html
@@ -14,7 +14,7 @@
         <!-- The Wick Engine build is injected here during the engine build step (see gulpfile.js) -->
         <script>
             /*Wick Engine https://github.com/Wicklets/wick-engine*/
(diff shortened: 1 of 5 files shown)


BaronAWC:
was able to test it myself and with the provided test file. though, i wonder if there's a way to give a more useful line number for the error...

image

Alani25 (approved):
Everything seems to work as intended!

PR #102, opened 2026-01-05
Node 24 support

bogoddr:
## Description

Wick Editor and Candlestick currently depend on Node 14 for development and creating builds. **Node 14 reached End-of-Life (EOL) on April 30, 2023, meaning it no longer receives security updates, bug fixes, or support.** https://nodejs.org/en/download/archive/v14.21.3

This PR addresses this issue by upgrading all dependencies to versions that support Node 24.

We no longer have any critical security issue warnings from `npm audit`! This is down from 22 critical issues using Node 14 and the old dependencies. 

## Current status

The editor builds successfully using node 24 on Ubuntu, MacOS, and Windows. All engine unit tests are passing (except for the few that have been broken for a long time).

Some additional testing of general editor functionality would be helpful here also!

## Additional Notes

- I removed electon and unused/outdated native+desktop build config + files
- I removed python installation in the readme since we no longer use `node-sass` (python isn't needed to build anymore)
- removed both of these unused dependencies:
  - "14": "^3.1.6"
  - "fs": "0.0.1-security
- Replaced deprecated node-sass with modern [sass implementation](https://github.com/sass/dart-sass)
- bumped "react-scripts" "2.0.5" -> "^5.0.1"
  - looks like create-react-app is deprecated. https://github.com/facebook/create-react-app/releases/tag/v5.0.1 
  - nowadays it looks like people are using Vite https://www.reddit.com/r/reactjs/comments/12x6k9x/it_looks_like_createreactapp_is_dead_what_should/
  - Long term, a migration to a more build tooling is needed (and could make desktop app builds easier) but it would be a signifigant effort, I'd say staying on create-react-app is fine for now.
- Added .svgrrc.js -- fixes SVG imports so they work as JSX imports in react-scripts 5.0.1 https://stackoverflow.com/questions/59820954/syntaxerror-unknown-namespace-tags-are-not-supported-by-default
- Added jsconfig.json -- react-scripts 5.0.1 breaks importing from "Editor/" which used to work because paths implicitly started with src/
  - This fixes that issue without having to update every single import in the project.
  - https://www.typescriptlang.org/tsconfig/#baseUrl
 - Bumped `gh-pages` to >5.0.0 (resolves https://github.com/advisories/GHSA-8mmm-9v2q-x3f9)


Alani25 (changes requested):
## Tauri Apps: Mac + Windows Build

**For mac built** it seems when the editor loads the inspector and main canvas + timeline aren't rendered at first. They only appear after a simple *window resize*.
image
(^ after resizing the window it looks like the normal editor and works normally ofc)

and this next thing is on **both windows and mac** — when you try dragging assets onto the canvas, the file drop overlay doesn't appear and the assets aren't added (same for dropping files)
<sup>This might also be on linux build though I haven't tested there</sup>
image
image

I believe these are most of the issues that I caught and think are related to the packages updates. Those asides, most things seem to work as intended, well done!

bogoddr:
Fixed file drag on drop on Windows per: https://github.com/tauri-apps/tauri/issues/14373

Haven't tested the fix on macOS yet, though

bogoddr:
Found a hacky way to solve the macos blank window bug! I'm not actually sure why the issue is happening though, need to dig deeper into how Tauri handles creating the app window.

bogoddr:
Resolved the last critical security issue, but we'll need to verify that deployment to github pages still works. (I bumped gh-pages from 2.0.1 to 5.0.0, I assume there have been breaking changes)

Alani25 (approved):

BaronAWC (changes requested):
tested and didn't see anything break, though the terminal is quite unhappy and is filled with warnings and deprecation notices.

there are 30-something warnings about failing to parse a source map for the node module "console-feed". after looking at https://github.com/samdenty/console-feed/issues/116 and https://github.com/facebook/create-react-app/discussions/11767, i think the simplest way to get rid of the warnings is to go to `package.json` and add to the `start` script:
`"start": "GENERATE_SOURCEMAP=false react-scripts start",`
but it looks like this can have some undesirable side effects for development, according to people in the 2nd link's conversation.

all the other warnings look like deprecations in various `.scss` files. while i don't see anything broken in the build, i do think it's a good idea to clear those out.

InternetAstronaut:
> all the other warnings look like deprecations in various `.scss` files. while i don't see anything broken in the build, i do think it's a good idea to clear those out.

There's a single SCSS function used in almost all the Sass part of the codebase that is deprecated.

We can choose to continue using it, but I think we _could_ fix this with another PR.

InternetAstronaut:
> There's a single SCSS function used in almost all the Sass part of the codebase that is deprecated.

I did some research, and the function is `darken()`. They deprecated it because it was global, and used a fixed step that doesn't correctly work with _some_ color spaces.

There's an easy fix, which is using the oficial Sass migrator to replace `darken()` and `lighten()` calls with the modern alternatives included in the color package.

```
npm install -g sass-migrator
sass-migrator module --migrate-deps STYLESHEET
```

We will need to migrate **EVERY** Sass file in the codebase, which (thankfully) just includes the utility components. (maybe _some_ modals have SCSS code?)

InternetAstronaut:
I recommend fast-tracking these changes, as **MacOS 28** (the version after Silk Road) will finalize the **transition to Apple Silicon**, **removing Rosetta 2 from MacOS permanently**.

_Since nobody here is going to rebuild, repackage and resign NodeJS 14 for Apple Silicon, it is simply better to fully move on to the more modern NodeJS 24._

BaronAWC:
Ah yeah I did kinda forget about the whole Rosetta 2 thing.

We're planning to merge this after the 1.0.2 release (which should be before July), so I will give this a review right after that.

.gitignore
```diff
@@ -10,9 +10,7 @@
 
 # production
 /build
-/build-ios/WebViewExample/local_html
```
BaronAWC, line 13: The only feedback I have would be that the gitignore doesn't seem to be properly ignoring the `dist` files.

For context/as a reminder, we originally did this in [this PR](https://github.com/Candlestickers/Candlestick/pull/127) because we wanted to eliminate merge conflicts, and these files are just auto-built so we don't need them to be tracked by git.

BaronAWC (changes requested):
I didn't test for long but nothing looked visibly broken, so that's good. After 1.0.2 I may want to do a more rigorous test by running some projects and stuff, and then it's basically good to go other than the one comment.

.gitignore
Alani25: added to gitignore and removed from cache

BaronAWC (approved):

PR #103, opened 2026-01-30
Updates - Gradient color picker

StickmanRed:
Does not need to be merged while the Node 24 updates are ongoing (excited for the upgrade!)

### Fixes
- Crashes when "Stop Offset" given values beyond [0,1] range
- Color picker pop-up displaced when switching from solid to gradient and vice versa
- Canvas tool and pop-up gradient slider don't match when switching from solid to gradient

### Changes
- Color picker inputs made smaller
- Cursor switched to "gradient" `CURSOR_GRAD` when hovering over gradient GUI (build engine to view this change)


bogoddr:
Looks good😎

PR #104, opened 2026-01-30
Feature - Clip skewing and tweens

StickmanRed:
### Changes
- Added `skew` property to `Wick.Transformation`, `Wick.Clip`, and `Wick.Tween`.
- Added `fromMatrix` and `toMatrix` functions to `Wick.Transformation`.
  - These functions use [Ruffle code](https://github.com/ruffle-rs/ruffle/blob/master/core/src/display_object.rs#L492-L499) as a reference. Since Ruffle is designed to replicate Flash projects, and the skew tweens act visually similar to Adobe classic tweens, I decided to refer to the above.
- Replaced instances where `Wick.Clip` uses or sets `paper.Matrix` transformations.

Requires an engine build to test.
Currently, this does not add a `skew` property to `Wick.Selection`, so clip skews cannot be changed in the Inspector. The selection box design isn't currently equipped to handle skew.

### DIFF
--- a/engine/src/Transformation.js
+++ b/engine/src/Transformation.js
@@ -26,6 +26,7 @@ Wick.Transformation = class {
      * @param {number} scaleX - The amount of scaling on the x-axis
      * @param {number} scaleY - The amount of scaling on the y-axis
      * @param {number} rotation - Rotation, in degrees
+     * @param {number} skew - Skew, in degrees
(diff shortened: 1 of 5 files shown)


InternetAstronaut:
@StickmanRed, why did you close this? There is strong community support and need for this feature.

hobbsythe6th:
@StickmanRed did successfully implement this on the fork, so I guess we'll get a new PR for clip skewing

PR #105, opened 2026-02-01
fix: Breaking apart parent clip makes child clips undeletable

StickmanRed:
Fixes [clip bug](https://forum.wickeditor.com/t/scratch-like-cursor/18102/52)

The bug was caused by `_willBeRemoved` being set to `true` by `clip.remove()` in `Wick.Frame.addClip`. Bugfix tested through console on Candlestick.

Requires engine build to test

### DIFF
--- a/engine/src/base/Frame.js
+++ b/engine/src/base/Frame.js
@@ -409,6 +409,7 @@ Wick.Frame = class extends Wick.Tickable {
         }
 
         this.addChild(clip);
+        clip._willBeRemoved = false; // Fixes a bug where breaking apart a parent clip causes child clips to be undeletable
 
         // Pre-render the clip's frames
         // (this fixes an issue where clips created from ClipAssets would be "missing" frames)


bogoddr:
Ooh I ran into this during the gamejam haha. Thanks for finding a fix

Do you still have the code you used to verify this fix? I'd like to add it to the engine test suite

StickmanRed:
Not really, I just used the inspector "Break Apart" action and manually deleted the child clips... this code could be used to verify it though
```javascript
project.activeFrame.clips[0].breakApart();
let clip1 = project.activeFrame.clips[0];
clip1.remove();
// clip1.parent should equal null
```
also, about the testing, would it be correct to add the bugfix test code after [this line in test.Wick.Clip.js](https://github.com/Candlestickers/Candlestick/blob/7cb0ac4cd0e1729e4d285f075724904218d64a37/engine/tests/test.Wick.Clip.js#L370)?

bogoddr:
Yup, I'd add the test there with a comment above it like

`// Check that child clips are deletable after break apart (https://github.com/Candlestickers/Candlestick/pull/105)`

I'm not sure if running the tests still works on node14, but they are working on the node 24 branch with new instructions: https://github.com/Candlestickers/Candlestick/pull/102/changes/d4797dd20e2080caa25dc94a9ecd65e7cda5eef1

PR #107, opened 2026-02-04
fix: Objects keep moving when Shift+Direction pressed

StickmanRed:
The bug was reported on [the forums](https://forum.wickeditor.com/t/objects-can-continue-moving-on-their-own-after-pressing-shift-arrow-keys/24022). This fix clears the previous key repeat intervals (if any) before setting a new repeat.


bogoddr:
Looks good!

PR #116, opened 2026-03-26
fix clip borders showing when they shouldn't

BaronAWC:
at some point, the logic for hiding the clip borders was commented out. this pr just uncomments and fixes the if statement.


StickmanRed:
Tested through console on Candlestick test page. The code solves bug both in editor and GIF export

PR #119, opened 2026-05-01
Update Readme and package.json EDIT: also remove this annoying error that keeps popping up at local runtime

hobbsythe6th:
I've linked Stickmanred's tutorial in the readme for Windows users, and I've removed the build-engine-windows script since shx can handle the Mac commands. Also, every other script with Mac commands (the copy-pdf-worker and prerelease-deploy scripts are two) has been updated with shx.


BaronAWC (approved):

Alani25 (approved):
Thanks for doing this — the `shx` changes make sense to me, and I like the switch from `;` to `&&` since now if a part of the command chain fails it stops at that point 

All your changes make perfect sense
And the additional link in the read me to stickmanred's window version of the installation process is the cherry on top. Thank you, PR approved 🫡

PR #123, opened 2026-05-06
Make toasts less obstructive

### LINKED ISSUE #113: adjust stylization of editor toasts
**Is your feature request related to a problem? Please describe.**
For various reasons, the editor's toasts get in the way of my workflow.
- they are absolutely huge
- they stay on-screen for quite a while, and their transition in and out are pretty slow (especially transition out)
  - while they are transitioning out, they still occupy space and the other toasts wait for it to be fully removed before snapping to their new positions
- they partly cover the top right buttons, like the save button
- some things, like variables having invalid names, really shouldn't be toasts at all

**Describe the solution you'd like**
The toasts should be much smaller, transition faster, and not cover too many buttons. Ideally, the warning about invalid variable names should be in the actual variable name field.

BaronAWC:
I have shrunk the toasts by quite a bit since that's just what I prefer, but it's definitely up for debate whether I went too small.

I also moved them to the bottom right, because I think it's more acceptable to cover the asset library than the top of the inspector and the save/settings/etc buttons. Additionally, this makes it so that when older toasts at the top slide away, it doesn't awkwardly shove the rest of the toasts up. (It does the weird shoving when a new toast is displayed instead, but I think it's more manageable for my eyes, since at least it catches my attention when there's something new.)

before: (note how the toast is blocking all the buttons in the top right and the name field in the inspector)
image

after:
image

Last few things I'm thinking about:
- adding in the progress bar for when a toast auto-closes
- speeding up the open/close animation
- Making all toasts auto-close after the same amount of time (right now it seems like toasts can be 3, 5, or 7 seconds)

### DIFF
--- a/src/Editor/Editor.jsx
+++ b/src/Editor/Editor.jsx
@@ -853,12 +853,12 @@ class Editor extends EditorCore {
 
         // Default options for the toast:
         let defaultOptions = {
-            position: "top-right",
+            position: "bottom-right",
             autoClose: 3000,
             hideProgressBar: true,
             closeOnClick: true,
             pauseOnHover: true,
-            draggable: true,
+            draggable: false,
             className: (type + '-toast-background'),
             bodyClassName: (type + '-toast-body'),
             progressClassName: (type + '-toast-progress'),
--- a/src/Editor/_editor.scss
+++ b/src/Editor/_editor.scss
@@ -102,15 +102,16 @@ body {
 // Toasts
 .Toastify__toast {
   font-family: 'Nunito Sans';
-  font-weight: 800;
-  font-size: 24px;
+  font-weight: normal;
+  font-size: 16px;
   border-radius: 3px;
+  min-height: 0;
 }
 
 .Toastify__toast button{
   font-family: 'Nunito Sans';
-  font-weight: 800;
-  font-size: 24px;
+  font-weight: normal;
+  font-size: 16px;
 }
 
 .info-toast-background {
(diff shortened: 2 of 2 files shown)


JMii63:
i have genuinely no idea what a toast is in animation, so it was absolutely hilarious seeing "make toasts less obstructive" like girlie are we being too loud with our toasts? (/j)

anyways this looks pretty good

BaronAWC:
Maybe an image would have been useful lol

Toasts are the notification boxes that appear on the top right (now bottom right in this PR) when you do things like uploading an asset, saving/exporting, or if you give an invalid name to a frame, clip, etc. I'd always thought they got in the way of important buttons, so this was my attempt to make them block fewer and less important things.

JMii63:
> Toasts are the notification boxes that appear on the top right (now bottom right in this PR) when you do things like uploading an asset, saving/exporting, or if you give an invalid name to a frame, clip, etc. I'd always thought they got in the way of important buttons, so this was my attempt to make them block fewer and less important things.

ohhhhhh I see yeah this makes sense

InternetAstronaut:
I think that putting the toast in the upper-middle or the lower-middle is a better option.

---

Lower Middle

---

image

Alani25:
While I do agree with UI changes to the toasts, that can come in a second PR dedicated to re-stylizing the notifications to make them more appealing and the sort. 
For now I'll give my approval for the changes here

@BaronAWC feel free to merge when you think it's right time

PR #132, opened 2026-06-03
Fix brush tool and fill bucket bugs on browser zoom

### LINKED ISSUE #95: [BUG] cant draw in corner
**Describe the bug**

cant draw in corner

**To reproduce**

it just happens just refresh a few times 

**Expected behavior**

i should be able to draw there

**Screenshots**

https://github.com/user-attachments/assets/eec1c884-2583-45f6-9e85-0ad1ff37b024

**Editor version**

yes it happens there too but i hope you guys can fix it 

**Device**
-Lenovo Yoga 11e 6th gen
-Windows 10 (it looks like windows 7 but its not https://reunion7.com/ )
-Superminium
-1.0.1

**Do you have a suggested solution to this issue?**

sorry no

**Additional context**

nothing

StickmanRed:
This resolves issue #95 and hopefully issue [Weird UI Glitch](https://discord.com/channels/1406876763920797717/1407064965029826692). For the brush tool, zooming out messes up the size attributes of Wick's `canvas` element: the width/height *attributes* of the canvas become scaled down from the true width and height. But the brush tool's `canvas` uses the smaller width/height attributes as it true size, which is why the brush strokes keep getting cut off.

For the fill bucket, I have no idea why the solution works

### DIFF
--- a/engine/src/tools/Brush.js
+++ b/engine/src/tools/Brush.js
@@ -378,10 +378,11 @@ Wick.Tools.Brush = class extends Wick.Tool {
             this.paper.view._element.parentElement.appendChild(this.croquisDOMElement);
         }
 
-        // Update croquis element canvas size
-        if(this.croquis.getCanvasWidth() !== this.paper.view._viewSize.width ||
-           this.croquis.getCanvasHeight() !== this.paper.view._viewSize.height) {
-            this.croquis.setCanvasSize(this.paper.view._viewSize.width, this.paper.view._viewSize.height);
+        // use the CSS pixels size (viewSize) rather than trying to predict with device pixels (element.width) — they change when browser zoom ≠ 100% -H.A.
+        var targetW = Math.round(this.paper.view.viewSize.width);
+        var targetH = Math.round(this.paper.view.viewSize.height);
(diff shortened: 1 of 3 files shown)


BaronAWC:
When using the brush, the size of the cursor does not properly scale with what you see being drawn on the screen. This is because the cursor's "zoom" is always at 100%, rather than whatever percent the tab is at.

I found this decade-old repo (https://github.com/tombigel/detect-zoom) that actually still works, which we could use to scale the brush/eraser cursor accordingly. Alternatively we can change the cursor to be drawn directly on the website instead of using the cursor (which could arguably be better, so that we can work around the maximum cursor size).

There are 2 more bugs I noticed that come from Wick, those being 1) if you zoomed in or out during the current session, the eraser gets weirdly offset/zoomed before doing the erase and 2) if you zoom in during the session, the resolution of the canvas tanks. both of these get fixed by reloading for the most part, the exception is that for the second bug, the raster brushstroke still looks a bit more blurry.

Didn't notice any *new* bugs from this though, so that's good. If these issues are easy to fix we can do them, but probably no need to hold back 1.0.2 for these.

BaronAWC (approved):

PR #137, opened 2026-06-16
Drag & Drop PDFs, MP4s, and COPY PASTE image files

Alani25:
You can drag and drop PDFs & MP4s to open them the same way you can open a wick file by dragging it into the editor, which is sometimes far faster than needing to manually click the "open" button then searching for the file to open.

Also, you can now copy-paste images directly from your clipboard to the editor.
Pasted images will automatically be added to the asset library and placed where the cursor is located. The added image asset is then automatically "copied" so that the next time the user pastes, it pastes the same image from the asset library rather than re-importing it.
When the user copies an object inside of the project, that object is then pasted rather than the image. If the user then copies an image, that image is what's to be pasted until another image/ wick object is copied.

Whatever was copied last is what gets pasted.
How can the editor tell what was copied last?

Whenever the user uses the paste or copy feature in candlestick, the current clipboard is saved within the window's local storage as a stamp "finger print." If the current clipboard has an image in it, but that image doesn't match the saved fingerprint, then we upload the image into the asset library then _copy_ the image wick object that is added to the project from that copy-paste. That way, the next time the user pastes, it'd paste the uploaded image rather than upload a new image. When the copy feature is used to copy wick elements, we fingerprint the current clipboard to make sure it's ignored. If the clipboard changes than what was stamped, then we know that the user copied an image after the copied wick object. The reason i save the fingerprint on the window's local storage is to make sure that users are still able to use copy/ paste features between multiple projects in multiple windows without issue.

It may be a bit confusing to explain it, but to put it simply this feature should work as "expected." 
Testing would be appreciated.

### DIFF
--- a/src/Editor/Editor.jsx
+++ b/src/Editor/Editor.jsx
@@ -350,6 +350,16 @@ class Editor extends EditorCore {
         };
         document.addEventListener('mousemove', this._mouseMoveHandler);
 
+        // Paste event listener — fires because we no longer call e.preventDefault() on the
+        // Cmd+V keydown for 'paste'. This gives us clipboardData with real File objects
+        // (original filenames, Finder file copies, Discord/Figma images, etc.)
+        this._pasteHandler = (e) => {
+            if (e.target.tagName === 'INPUT' || e.target.tagName === 'TEXTAREA') return;
+            e.preventDefault();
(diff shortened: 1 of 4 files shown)


InternetAstronaut (approved):
LGTM

PR #138, opened 2026-06-19
Image smoothing as a setting

### LINKED ISSUE #3: Disable image smoothing (setting option)
Image smoothing really kills pixel art, however I can understand why some people might find it useful.
This is why we should consider making image smoothing a settings option. 

To compare, here's an image from wick of an important pixel image with and without image smoothing:
image
It makes a huge difference with pixel art. 

You can disable image smoothing through Wick, code, but it's kinda annoying.
Check out `@SomeoneElse`'s thread on Wick forums for more:
https://forum.wickeditor.com/t/removing-image-smoothing/8582?u=hamzah_alani

hobbsythe6th:
I've added image smoothing as a setting, and it works I think. I have very little experience in testing for bugs, so some may be found. This is also a global feature, meaning that as of now you can't (or aren't supposed to) have image smoothing on for one and off for the rest using the setting. Technically the on/off script only runs if you change the setting, and you can still turn smoothing on/off through code, but it's a bit less likely a use case to me so I left it alone.
I don't know if you want to have smoothing in the inspector or not, but it's a bit harder to do that and the global setting, but I'm working (sadly very slowly) on it. That said, I think this is fine for most use cases.
Also I accidentally put the commits from the other pull I made on this pull, and idk what to do. Merge that one first maybe?
Closes #3


BaronAWC:
Was trying it just now and it seems like the image always reverts back to smoothening after any editor interaction.

I'm not too sure myself on whether a global or image-specific setting is better (it's probably worth a full discussion to decide), but I think it should be a project setting rather than an editor setting at the very least.

InternetAstronaut:
> the image always reverts back to smoothening after any editor interaction

I think the issue is that the option just goes child-by-child changing the path's smoothing property, for this to work as intended it must applied to images being created.

I would also argue that there isn't any _good reason to implement such a difficult thing_, There haven't been many pixel art animations/games using Wick, and this just feels like _implementing for the sake of implementing_.

BaronAWC:
I can say for myself that I've made one pixel art game semi-recently and one pixel art animation waaaaaay in the past. 1x-resolution pixel art would show up blurry, and while I was able to just solve it by exporting at 10x resolution, I do think that an anti-alias option is helpful. It would mean that you can make your pixel art as zoomed in as you want while still looking crisp.

Though thinking about it more, there isn't really a compelling reason to make this an object-specific setting, so a project-wide toggle is probably better.

This will most likely be a 1.1.0 kind of deal so no there's no rush.

hobbsythe6th:
Do you have any idea where the bug might be? because I have no ideas at all

BaronAWC:
I'll be busy for a while so I can't look too closely at this for at least a week, and honestly I don't really know off the top of my head either, but I can try tackle it next week.

hobbsythe6th:
I think there's two ways to fix this right now: put the on/off code into a loop that runs every time the project renders (which was my v1 for this pr) and the other way is to find what reverts the smoothing to original and disable it (i suspect it is involved with the projectDidChange function but I'm not sure)

engine/src/ToolSettings.js
```diff
@@ -195,6 +197,13 @@ Wick.ToolSettings = class {
         } else {
             localforage.setItem(this.getStorageKey(name), value);
         }
+        if (setting.name === "imageSmoothing"){
+            this.project.getAssets("Image").forEach(asset => {
+                asset.getInstances().forEach(img => {
+                    img.view.item.smoothing = setting.value
```
InternetAstronaut: this doesn't actively apply to newly created GUIElements. You can change in the GUIElement base code to look for this setting.

InternetAstronaut (changes requested):

StickmanRed:
> I think there's two ways to fix this right now: put the on/off code into a loop that runs every time the project renders (which was my v1 for this pr) and the other way is to find what reverts the smoothing to original and disable it

I think the problem is, whenever the project renders, [this function (View.Path.js)](https://github.com/Candlestickers/Candlestick/blob/baa008571f57b74d8bdd3b69c4082e06613b6341/engine/src/view/View.Path.js#L44) rebuilds the image from scratch, which reverts the image smoothing setting. So we can move the on/off code to the render function:

```javascript
    render () {
        if(!this.model.json) {
            console.warn('Path ' + this.model.uuid + ' is missing path JSON.');
            return;
        }

        this.importJSON(this.model.json);

        // Apply image smoothing setting
        if(this.model.pathType === 'image') {
            this.item.smoothing = this.model.project.toolSettings.getSetting('imageSmoothing');
        }

        // Apply onion skin style if Needed
        // ...
```

InternetAstronaut:
> > I think there's two ways to fix this right now: put the on/off code into a loop that runs every time the project renders (which was my v1 for this pr) and the other way is to find what reverts the smoothing to original and disable it
> 
> I think the problem is, whenever the project renders, [this function (View.Path.js)](https://github.com/Candlestickers/Candlestick/blob/baa008571f57b74d8bdd3b69c4082e06613b6341/engine/src/view/View.Path.js#L44) rebuilds the image from scratch, which reverts the image smoothing setting. So we can move the on/off code to the render function:
> 
> ```js
>     render () {
>         if(!this.model.json) {
>             console.warn('Path ' + this.model.uuid + ' is missing path JSON.');
>             return;
>         }
> 
>         this.importJSON(this.model.json);
> 
>         // Apply image smoothing setting
>         if(this.model.pathType === 'image') {
>             this.item.smoothing = this.model.project.toolSettings.getSetting('imageSmoothing');
>         }
> 
>         // Apply onion skin style if Needed
>         // ...
> ```

LGTM!

hobbsythe6th:
for some reason this just makes me worry about the inefficiency, even though I know it's the best way to do it. Thanks so much for the fix! I'll update it soon

InternetAstronaut:
> for some reason this just makes me worry about the inefficiency, even though I know it's the best way to do it. Thanks so much for the fix! I'll update it soon

Yes, yes, this method can be inefficient since we're writing to memory every single render with this. Writing to memory _may be slow_, because writes that take a _few microseconds_ can **stack to dramatic performance losses** (e.g.: 100 miliseconds), specially when the changed data isn't in the CPU cache.

StickmanRed:
> Though thinking about it more, there isn't really a compelling reason to make this an object-specific setting, so a project-wide toggle is probably better.

I can work on moving the setting to a project setting and add the changes on this branch, if it's okay with you @hobbsythe6th

hobbsythe6th:
I added a checkbox under the editor tab in the project settings to toggle it.

engine/src/base/Project.js
```diff
@@ -106,6 +106,7 @@ Wick.Project = class extends Wick.Base {
         this.activeTool = 'cursor';
 
         this._toolSettings = new Wick.ToolSettings();
+        this._toolSettings.project = this
```
InternetAstronaut: Is it correct to do this?

engine/src/view/View.Path.js
```diff
@@ -49,6 +49,10 @@ Wick.View.Path = class extends Wick.View {
 
         this.importJSON(this.model.json);
 
+        if(this.model.pathType == 'image') {
```
InternetAstronaut: This `if` clause gets called every frame, change it to a cached `this._isImagePath`

engine/src/view/View.Path.js
```diff
@@ -49,6 +49,10 @@ Wick.View.Path = class extends Wick.View {
...
+            this.item.smoothing = this.model.project.toolSettings.getSetting('imageSmoothing')
```
InternetAstronaut: Cache this also.

engine/src/ToolSettings.js
```diff
@@ -118,6 +122,8 @@ Wick.ToolSettings = class {
         this._settings = {};
         this._onSettingsChangedCallback = () => {};
 
+        this.project = null
```
InternetAstronaut: Why did this get added, and is it safe to other unchanged code?

engine/src/base/Project.js
InternetAstronaut: Isn't it simpler to just ask the project in the tool settings' constructor?

engine/src/base/Project.js
```diff
@@ -106,6 +106,7 @@ Wick.Project = class extends Wick.Base {
         this.activeTool = 'cursor';
 
         this._toolSettings = new Wick.ToolSettings();
```
InternetAstronaut, line 110:
```suggestion
        this._toolSettings = new Wick.ToolSettings(this);
```
[Example addressing this.](https://github.com/Candlestickers/Candlestick/pull/138/changes#r3832953480)

engine/src/ToolSettings.js
InternetAstronaut:
```suggestion
        this.project = project || null;
```

engine/src/ToolSettings.js
InternetAstronaut, line 117:
```suggestion
    constructor (project) {
```

InternetAstronaut (changes requested):
These changes are confusing me, why are we setting the project for `ToolSettings` as null, just to reset in the project?

Also doing a string equation and method call to tool settings every frame in every path?

StickmanRed:
> I added a checkbox under the editor tab in the project settings to toggle it.

I meant make `imageSmoothing` a property of the project `project.imageSmoothing`, as opposed to just a tool setting `project.toolSettings.getSetting('imageSmoothing')`, so that it applies to exported `.html` or `.zip` projects. If it's okay, I can push some changes to this branch that implement this

hobbsythe6th:
Yes, I'm fine with that. This branch is kinda abandoned right now on my end.

engine/src/base/Project.js
hobbsythe6th: I don't remember adding this, I'll change it

engine/src/view/View.Path.js
hobbsythe6th: I don't understand, how should it be cached?

hobbsythe6th:
Updated it.

engine/src/view/View.Path.js
InternetAstronaut: Well, we only change by tool settings, so we can push a boolean in the project (`project._imageSmoothing`) which would be faster.

StickmanRed: Project now uses an `imageSmoothing` property, resolving

engine/src/base/Path.js
InternetAstronaut, line 1: Revert these changes, you made it so the caching is gone, and now checking path type takes two instance checks.

engine/src/base/Project.js
```diff
@@ -38,6 +39,7 @@ Wick.Project = class extends Wick.Base {
         this._height = args.height || 480;
         this._framerate = args.framerate || 12;
         this._backgroundColor = args.backgroundColor || new Wick.Color('#ffffff');
+        this._imageSmoothing = (typeof args.imageSmoothing === 'boolean') ? args.imageSmoothing : true;
```
InternetAstronaut:
```suggestion
this._imageSmoothing = !!args.imageSmoothing;
```

engine/src/base/Project.js
```diff
@@ -347,6 +351,19 @@ Wick.Project = class extends Wick.Base {
...
+     * Whether to apply image smoothing to project's images.
+     * @type {boolean}
+     */
+    get imageSmoothing() {
+        return this._imageSmoothing;
+    }
+
+    set imageSmoothing(imageSmoothing) {
+        // Apply image smoothing for old projects
+        this._imageSmoothing = (typeof imageSmoothing === 'boolean') ? imageSmoothing : true;
```
InternetAstronaut:
```suggestion
this._imageSmoothing = !!imageSmoothing;
```

InternetAstronaut (changes requested):

engine/src/base/Path.js
StickmanRed: I think `pathType` caching would belong better in a separate PR, since this PR is dedicated to image smoothing. Instead, I replaced the `pathType` accessor with the instance check against `paper.Raster`

engine/src/base/Project.js
StickmanRed: I admit the `typeof` stuff does look kinda weird, I used it so previous Wick projects with `imageSmoothing = undefined` default to `imageSmoothing = true` for backwards compatibility. But maybe it isn't necessary, as users can quickly reenable image smoothing; what do you think?

InternetAstronaut (approved):

Issue #139, opened 2026-06-20
Add An Export To WickObj button


TSFBCE24RhythmHeaveners:
**Is your feature request related to a problem? Please describe.**
Yes, for those who use keyboardless devices

**Describe the solution you'd like**
An export selected clip to WickObj button

**Describe alternatives you've considered**
Its own keyboard shortcut.

**Additional context**
Great for Android device users.

InternetAstronaut:
Already possible. `WICKOBJ` can contain any object that extends `Wick.Base`.

We can add the export inside the code that already does this.

TSFBCE24RhythmHeaveners:
> Already possible. `WICKOBJ` can contain any object that extends `Wick.Base`.
> 
> We can add the export inside the code that already does this.

I said a physical button.

PR #140, opened 2026-06-21
Add ability to export buttons and paths as wick object files

InternetAstronaut:
This PR adds #139. I don't think I need to clarify any more what and how this change works.

### DIFF
--- a/src/Editor/EditorCore.jsx
+++ b/src/Editor/EditorCore.jsx
@@ -2055,10 +2055,10 @@ class EditorCore extends Component {
     exportSelectedClip = () => {
         var clip = this.project.selection.getSelectedObject();
         if (!clip) return;
-        if (!(clip instanceof window.Wick.Clip)) return;
+        if (!([window.Wick.Clip, window.Wick.Button, window.Wick.Path].some((o) => clip instanceof o))) return;
 
         window.Wick.WickObjectFile.toWickObjectFile(clip, 'blob', file => {
-            window.saveFileFromWick(file, (clip.identifier || 'object'), '.wickobj');
+            window.saveFileFromWick(file, (clip.identifier || /*Clip::f87f8b7d-62c3-4576-a5bb-61ce87768ce9*/ `${clip.classname}::${clip.uuid}`), '.wickobj');
         });
     }
 
--- a/src/Editor/EditorCore.jsx
+++ b/src/Editor/EditorCore.jsx
@@ -2053,12 +2053,20 @@ class EditorCore extends Component {
     }
 
     exportSelectedClip = () => {
-        var clip = this.project.selection.getSelectedObject();
-        if (!clip) return;
-        if (!([window.Wick.Clip, window.Wick.Button, window.Wick.Path].some((o) => clip instanceof o))) return;
+        var object = this.project.selection.getSelectedObject();
+        if (!object) return;
+        if (!object instanceof Wick.Base || !(object instanceof Wick.Asset))) return;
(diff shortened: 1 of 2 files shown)


src/Editor/EditorCore.jsx
```diff
@@ -2055,10 +2055,10 @@ class EditorCore extends Component {
     exportSelectedClip = () => {
         var clip = this.project.selection.getSelectedObject();
```
BaronAWC: since this technically doesn't have to be a clip, i think renaming this variable to `object` makes more sense

src/Editor/EditorCore.jsx
```diff
@@ -2055,10 +2055,10 @@ class EditorCore extends Component {
...
         if (!clip) return;
-        if (!(clip instanceof window.Wick.Clip)) return;
+        if (!([window.Wick.Clip, window.Wick.Button, window.Wick.Path].some((o) => clip instanceof o))) return;
 
         window.Wick.WickObjectFile.toWickObjectFile(clip, 'blob', file => {
-            window.saveFileFromWick(file, (clip.identifier || 'object'), '.wickobj');
+            window.saveFileFromWick(file, (clip.identifier || /*Clip::f87f8b7d-62c3-4576-a5bb-61ce87768ce9*/ `${clip.classname}::${clip.uuid}`), '.wickobj');
```
BaronAWC: text objects are exportable, but i think it'd be nice if the classname was registered as `Text` instead of `Path`.

also, having the UUID and date of export doesn't seem that helpful, and all it does is make the filename extremely long. (especially the UUID - it doesn't help the user at all.) i'd say that unnamed objects are just named `clip.wickobj`, `button.wickobj`, `path.wickobj`, or `text.wickobj`. if the object has a name, use that instead.

BaronAWC: also the commented Clip::... can probably be removed

BaronAWC:
I've been thinking more about this and I'm wondering if it's actually a bad idea to export paths as a wickobj. If we work on getting the clips system closer to how symbols work in Flash, it might make more sense to only let clips (and buttons) be a wickobj, since only clips should have a parent/instance relationship.

BaronAWC:
Oh wait actually, this PR doesn't technically implement the feature request. It's specifically asking for a button in the inspector to export the object.

InternetAstronaut:
> Oh wait actually, this PR doesn't technically implement the feature request. It's specifically asking for a button in the inspector to export the object.

Yeah, my bad...

BaronAWC (commented):
forgot to actually submit this review

src/Editor/EditorCore.jsx
InternetAstronaut: > text objects are exportable, but i think it'd be nice if the classname was registered as `Text` instead of `Path`.

I also think text should be separate from paths, but that’s the way it was coded and it’ll break old files if we separate the two. I think we can change the `classname` to “Text” in the better, since only internal code reads that

InternetAstronaut: > also the commented Clip::... can probably be removed

I know, I just put it to get a reference.

InternetAstronaut: > also, having the UUID and date of export doesn't seem that helpful, and all it does is make the filename extremely long. (especially the UUID - it doesn't help the user at all.) i'd say that unnamed objects are just named `clip.wickobj`, `button.wickobj`, `path.wickobj`, or `text.wickobj`. if the object has a name, use that instead.

That also seems like something good.

PR #144, opened 2026-06-21
fixed problem where objects could be added with incorrect identifiers

### LINKED ISSUE #142: [BUG] when adding a wickobj to a canvas, it bypasses the duplicate name detection
**Describe the bug**
It is possible to have multiple objects with the same name if they are added via wickobj.

**To reproduce**
1. upload a wickobj that has a name
2. add it to the canvas multiple times - they all have the same name

**Expected behavior**
When adding a wickobj, it appends `_copy` if a duplicate name is found.

**Editor version**
reproducible on all versions

InternetAstronaut:
This commit fixes #142. Go to the mentioned issue for clarification.

### DIFF
--- a/engine/src/base/Base.js
+++ b/engine/src/base/Base.js
@@ -512,6 +512,7 @@ Wick.Base = class {
      * @param {Wick.Base} child - the child to add.
      */
     addChild(child) {
+        if(!(child instanceof Wick.Base)) throw new TypeError("Child isn't a valid wick object.");
         var classname = child.classname;
 
         if (!this._children) {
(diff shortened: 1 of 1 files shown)


BaronAWC (changes requested):
Getting an error whenever I try to add a wickobject to a canvas.

Image

InternetAstronaut:
> Getting an error whenever I try to add a wickobject to a canvas.
> 
> Image

I think I know why. `_getUniqueIdentifiers()` recursively calls itself on the children of the `Wick.Base` object, so eventually the call stack will burst if there's too many items.

I think I'll make some quick changes!

InternetAstronaut:
Should be fixed with [commit 5c6d9e5](https://github.com/Candlestickers/Candlestick/pull/144/commits/5c6d9e51c9ad47390d401bea275a265c944b21ad).

kendrickroberts-dev:
wow

kendrickroberts-dev:
wow

PR #145, opened 2026-06-21
Fix brush tool breaking on layers with images or clips

StickmanRed:
When clips or images are in the layer, the boolean operation breaks and the brush tool stops working. This fix makes sure that each child object is a path before including it in the boolean operation.


InternetAstronaut (approved):
LGTM

BaronAWC (approved):

PR #149, opened 2026-06-28
ffmpeg upgrade

Alani25:
Our current development branch is on an old react version, iow webpack version 4, while the modern ffmpeg is an ES module distributed for webpack v5 specifically, so this update was branched off of @bogoddr's #102 (node v24)

The goal is to merge this back into Zrispo's branch as development is still at node v14 (to be updated in v1.0.3 once 1.0.2 is deployed).

Use this video on the Wick Editor Forums to compare the difference in MP4 export:
https://forum.wickeditor.com/t/nothing-shows-up-in-exported-video/24260
<sup>^ this specific file always results in broken results when exported, making it great for testing this</sup>

Current Candlestick MP4 exporter:
image

Updated exporter:
image

The results are _very noticeable_, to say the least.


BaronAWC (commented):
Tested by uploading various videos to Discord and viewing it on my Mac and iPhone, all seems good. Video quality is also *much* better. (Not sure if it's just my ears deceiving me, but I think the audio quality did go up a bit so that's a nice bonus. Unfortunately not all audio problems are solved, but that can come a bit later.)

I do see the engine/dist build files in here though, not sure if that matters though since I know in the Node 24 PR the gitignore shooould catch it?

I think it'd help if other people use other, non-Apple devices to test various animations, export on 1.0.1 and this branch, and test both of them in places like a native video player, Discord, etc. to see if the new one works as expected. If all of it goes well then I think this is all set.

Alani25:
whoops ignore that…
anyways yeh in regards to testing @BaronAWC I believe once we deploy this to the test editor we can point everyone with MP4 export problems there (as it's not a rare occurrence with the current state of the editor) and that should serve as a good test— which is why I think it'd be good idea to deploy this to development right after 1.0.2 is out. 
The longer it's in development the more likely people are to catch any issues.

PR #150, opened 2026-06-30
Add fix for vcam lagging behind during rendering

StickmanRed:
This fixes a vcam bug in exports where the vcam asset would be visible on the first frame. Tested locally by exporting a MP4, both video and audio works as expected.


InternetAstronaut (approved):
LGTM

Alani25:
All works great

Test:

https://github.com/user-attachments/assets/414bdb98-a05d-4bfd-ae98-c3165a7fac08

PR #152, opened 2026-06-30
Fix bugs with color picker and add QoL

### LINKED ISSUE #124: [BUG] If a path has a null color, gradient editor does not work
**Describe the bug**
Paths like brushes and pencils will have null border or fill colors. When trying to switch to gradient mode, it will briefly flash a faulty UI and revert to the solid color mode.

**To reproduce**
1. create a brushstroke
2. open the stroke color menu
3. try to switch to gradient mode

This also applies to the line and pencil tools with the fill color.

**Expected behavior**
The gradient mode works as usual. It should behave as if the color was black by default.

**Editor version**
Reproducible on the test version of Candlestick and the development build.

**Additional context**
The bug can be avoided by picking an arbitrary color in the solid color mode before switching to gradient.

BaronAWC:
Fixed gradient color-picking interactions with null colors and hitting undo while the picker is active.
Added the split solid/transparent color preview to all color pickers.
Added hover preview to creating a new stop inside the inspector.
Allowed clicking outside to exit the color picker.

### DIFF
--- a/engine/src/tools/Cursor.js
+++ b/engine/src/tools/Cursor.js
@@ -117,6 +117,7 @@ Wick.Tools.Cursor = class extends Wick.Tool {
 
                 if(stopIndex !== null) {
                     this._selection.selectedStopIndex = stopIndex;
+					widget._updateItems();
                     this.fireEvent({eventName: 'canvasModified', actionName: 'cursorSelectStop'});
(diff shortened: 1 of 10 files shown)


Alani25:
Looks good

PR #158, opened 2026-07-03
Add Tagging system

InternetAstronaut:
Adds a basic tagging system, with the following interface described in TypeScript:

```ts
declare class Wick.Base {
  declare public get tags(): string[]
  declare public set tags(tags: string[])
  declare public hasTag(tag: string): boolean
  declare public hasTags(...tags: string): boolean
  declare public hasAnyTags(...tags: string): boolean
  declare public addTag(tag: string)
  declare public addTags(...tag: string)
}
```


hobbsythe6th:
I did find tagging in the original Wick repo (branch 1.20), along with Inspector modifications. Maybe you could use it as a reference?

InternetAstronaut:
> I did find tagging in the original Wick repo (branch 1.20), along with Inspector modifications. Maybe you could use it as a reference?

Being very honest, I think this implementation is already very good, and I don't want to mess with the almost dying codebase (which I have little to no experience with) that is the editor.

InternetAstronaut:
> I did find tagging in the original Wick repo (branch 1.20), along with Inspector modifications. Maybe you could use it as a reference?

I have seen the original 1.20 tagging system, and I can say it was way more limited than this one.

The old system is Clip-exclusive, but also has the total tagbase of the project stored.

The inspector thing is also really simple to add, but I don't usually mess with the 'editor' part of Wick due to it requiring Node 14.

Maybe I'll try adding it later when this gets pushed.

PR #159, opened 2026-07-03
Alani copy paste bug

Alani25:
Fixes:
- Copy-paste erroring when clipboard permissions aren't accepted by default (null check)
- If copying invalid file type (ex: GIF), falling back to wick obj clipboard


StickmanRed (approved):
Tested this on mobile and the bug was fixed

PR #161, opened 2026-07-03
Possible fix: Color picker preview doesn't show correct color

StickmanRed:
I can't replicate this bug, but a possibility is that it may be caused by code not supported by older devices. This fix replaces `.toSorted` and `linear-gradient` with only one color stop.


BaronAWC (approved):
can confirm that the solid color is back for my older mac mini. next step would be to make sure the opaque half displays for older systems too. it's probably urgent enough to just push this right now and to fix the opaque thing later though, so this is good.

image

PR #166, opened 2026-07-05
settings save between sessions

### LINKED ISSUE #75: [BUG] "Editor" settings are not remembered between sessions
**Describe the bug**
the editor settings (which yeah no one really uses) get reset upon reloading the page.
Image

**To Reproduce**
1. change onion skinning style
2. reload and check onion skinning style

**Expected behavior**
editor settings are saved between sessions

keyboard shortcuts are saved, so there should be a way to save editor settings in a similar way

Alani25:
The shortcut settings already saved within the local forage but weren't being loaded properly. Fixed that in `EditorSettings.jsx`, and added the timeline settings as well (gap fill mode and frame size in bottom left).

To test: 
- Adjust at least one shortcut setting (ex: change preview from <kbd>k</kbd> to <kbd>enter</kbd>)
- Edit onion skinning settings
- Change gap fill mode and/ or frame size preference (bottom left)
Then reload the window, and test to see if your same settings/ preferences are still set.

### DIFF
--- a/engine/src/ToolSettings.js
+++ b/engine/src/ToolSettings.js
@@ -259,18 +259,9 @@ Wick.ToolSettings = class {
     loadSettingsFromLocalstorage () {
         Wick.ToolSettings.DEFAULT_SETTINGS.forEach(setting => {
             // Get stored tool setting if it exists.
-            localforage.getItem(this.getStorageKey(name)).then( (value) => {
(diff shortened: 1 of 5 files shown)


StickmanRed:
Tested locally and all three steps work. Great changes, I especially like that frame size is saved

PR #172, opened 2026-07-10
Add clips to asset library

hobbsythe6th:
Added a feature to add a clip to the asset library. Tested with buttons as well (but I don't know the first thing about testing so.. yeah) I've also fixed a little typo in the param for the wickobj export function, because the current param says it exports a Wick.Project :P


Alani25:
Few suggestions:
1. Add icon to the button
image

2. After asset is added to library, remove the "Add asset to library" button unless the clip's timeline is edited

3. Adding multiple assets to the asset library of the same name shouldn't be allowed— instead, there should be a count added at the end of the filename for each new asset that shares a same name.
image

4. When item has no name it might be better to give it a shorter name like "clip (#)" where # is the number. 

5. Adding clips with no name to asset library shoots out 3 notifications at a time. Even without a name, one notification should be enough (this is more of a picky suggestion and not necessary)
image

hobbsythe6th:
I tried adding an icon but it was SO HARD, so I didn't do it. I'll try getting the rest done soon, thanks for the feedback! Also, a question: would condensing the three toasts to one mean removing the 'importing files' toast? Since that touches the importing code and also it's basically unnecessary, because if there's an importing issue it'll throw an error anyway, so the in progress toast does nothing except tell people sometimes that the import process is really slow.

hobbsythe6th:
Clips are now numbered on adding to asset library, and clips can't be added to the asset library if they have an asset source. The 'importing files' and 'no identifier' toasts have been removed. Still working on making the 'add to library' button show up after editing the clip.

hobbsythe6th:
Update: now, if you change the project's focus by using the breadcrumbs button, the old focus is marked as something you 
can add to the asset library again. There are a few issues with this, such as the double-click-on-canvas to change focus not working, but in the case you want to re-add to the asset library it isn't too hard to enter the clip again and then use the breadcrumbs button to come back. I did try to make the double-click work, but it was behaving oddly so I left it alone. Also, just opening and then exiting the timeline of the clip makes the 'asseted' flag activate, but that's not too much of a bother.

hobbsythe6th:
Another update: if an imported asset has the same name as an existing one, it renames it with -copy. For some reason, you have to do the import twice if there's a duplicate, but I think it's fine for now.

Alani25:
> would condensing the three toasts to one mean removing the 'importing files' toast? Since that touches the importing code and also it's basically unnecessary, because if there's an importing issue it'll throw an error anyway, so the in progress toast does nothing except tell people sometimes that the import process is really slow.

For quick imports like adding a clip to the asset library, it's highly unnecessary since that takes a second and doesn't require any loading. For importing larger assets into the asset library, it's more necessary to let the user know it's being worked on, as it would take a minute. I'd advise adding an extra parameter to the function that's set to false by default for disabling or enabling that extra toast notification, and for importing assets you can disable it.

> I tried adding an icon but it was SO HARD, so I didn't do it.

If you're still stuck there I can help :+1:

Matter of fact, I see you've gotten decent progress done that I've yet to test— if you're okay with me collabing with your code, feel free to list all issues you're coming by and (when you're ready) give me permission to look at it and start doing changes

I was also considering turning this into more of an adobe-like "symbol" feature… in which case we'd need to track the original clip based off UUID and toss that into the asset library instead, while also having it update everywhere the asset is used— but for now that's just a thought (as there has been requests for a feature as such). 
If you think that's a good direction to go, I can help there as well

<sup>(extra note— I apologize for the late response it seems we're in opposite timezones)</sup>

hobbsythe6th:
Symbols are probably something for another PR imo. Also, I forgot to say that there's only one toast now since I swapped the createAssets() out for importFileAsAsset(). And yeah, I'm fine with collabing and if you could add an icon it would be really helpful, thanks! I'm quite confused about how to detect in every scenario if a clip has been edited, which is why it's a bit messy with the breadcrumbs button, but the only places it can really slow you down is when you have to edit a clip RIGHT after you make it an asset. But for the core importing, I think we're ok.

InternetAstronaut:
Symbols are simple to implement. We could just implement a `LiveClipAsset` that is both a `ClipAsset` and an active clip. After that, we can wire the properties (timeline & scripts) to the asset. Editing and looking into the timeline/scripts is literally just redirecting you to the asset.

hobbsythe6th:
Honestly that's a good idea @InternetAstronaut, my first thought for symbol implementation was to make all the clips sync their edits with the asset or re-add the asset on edit but this is much better

PR #173, opened 2026-07-14
Brush tool improvements

### LINKED ISSUE #18: New Brushes
Add in new brush modes.

See Nick's Wick Brushes repository:
https://github.com/Wicklets/WickBrush?tab=readme-ov-file

Demo:
https://wicklets.github.io/WickBrush/

This feature would be mainly beneficial to artists and animators.
[Suggestion by emailstudies on forums](https://forum.wickeditor.com/t/wickbrush-for-custom-brush-help/20361)

### LINKED ISSUE #64: [BUG] brush precision decreases with smaller brushes
**Describe the bug**
when using small brushes, whether with relative brush or not, the line weight is inconsistent and oversmoothened sometimes.

**To Reproduce**
1. use a very small brush and/or zoom out
2. draw a shape with the brush

**Expected behavior**
the brush stays reasonably precise regardless of size

**Screenshots**
as posted by Twisted in the wick editor discord
Image

**Desktop (please complete the following information):**
 - OS: [e.g. Windows, MacOS]
 - Browser [e.g. Chrome, Firefox]

StickmanRed:
### Features
- Improved brush settings UI in Inspector
- Brush shape options
- Brush spacing and resolution sliders
- Randomized brush offset and rotation options


InternetAstronaut (approved):
LGTM

BaronAWC (changes requested):
First of all I wanna say this is awesome, thanks for putting in all this effort. It works well but I wanna see if we can improve the experience. It's a long and scary list but I think these are all important to address.

# Clarity
- [x] "Reso" (resolution) doesn't really make sense as a name from the user's perspective. As devs we know why it's resolution, but it would be nice if resolution was renamed to something that everyone can understand. My suggestion would be "quality".
- [x] Spread should not have a button to toggle a slider between 0.01 and the max, it should just be a slider between 0 and the max. Also, it's called "scatter" in one place and "spread" in another, so let's make sure that the name is standardized.
- [x] We should consider removing some of the preset shapes and/or putting them in a dropdown. While they're nice to have, it creates the illusion of so many buttons that it's overwhelming. The plus sign also makes me think that it's a button to add a new brush shape, which is not how it works. I think that, at minimum, we should remove the rough, scatter, and plus. The dropdown is so that we can hide away all the buttons when we don't really need them.
- [x] The "delete brush" button should be grayed out and non-interactive if there is only one brush.
- [x] There are a bunch of buttons that should be moved or removed.
    - [x] In the brush selector, the "add brush" and "upload brush" buttons should be related. This could mean putting them in a dropdown where all the other brushes are, or 2 large buttons where "edit brush" is now.
    - [x] Also in the brush selector, "edit brush" and "delete brush" should be similar. In my opinion it's weird that the edit button is big while the delete button is small. They should probably go together at the bottom.
    - [x] The way that "delete brush" and "delete shape" are both trash can buttons in the same location is really confusing. The previous checkbox already has my suggestion for delete brush. For delete shape, it really shouldn't be in the same row of buttons as everything else since it serves a completely different function. I don't know where exactly it should go... but not where it is now.
    - [x] When editing a brush, the pressure, relative size, and mode buttons should not be there. They are accessible when outside of the brush editor because they can change at any time without "altering the fundamental brush". Therefore, the brush editor should not show these, as it's only for the "important" settings.

# Customizability
- [x] Some shapes like the thin rectangle rotate to follow the cursor as it moves, but other brushes like the square don't. This makes me think that the rotation thing is defined in the brush shape, which doesn't make sense. I think there should be 3 rotation options: set offset angle (default is zero), follow mouse (with offset angle), and randomize angle.

# Visual Improvements
- [x] When dragging the sliders, the preview flickers quite a bit, which is distracting. Hopefully there's a way to remove the flickering?
- [ ] I found some inconsistencies in the shapes. The sparkle in the preview is in an "X" shape, but it's in a "+" shape everywhere else. The moon looks different in all 3 places: on the canvas it's closest to a "C" shape, on the button it's not as sharp, and in the preview it's cut weirdly.
- [ ] When using a big brush size, the brush path before releasing the mouse is pretty low resolution. Is it possible to increase the resolution so that it looks nicer?
- [x] The triangle shape is not equilateral... I don't know, that kinda bugs me. (If we fix this, make the centroid/center of mass the middle of the shape if that's even possible)

# Safeguards
- [ ] We should have preset brushes that cannot be significantly modified or deleted. For the standard brush, the only customizable settings should what we offer in the current version of Candlestick, plus maybe the resolution/smoothing.
- [ ] It's a little dangerous to be able to delete some of the fundamental shapes, especially the circle. I toyed with deleting stuff and I had to clear my cookies to bring back the default shapes. I'd say that the default shapes should not be allowed to be deleted, or at bare minimum the circle.
- [x] There should be a cancel button to exit the brush config without saving it.

Alani25:
> "Reso" (resolution) doesn't really make sense as a name from the user's perspective. As devs we know why it's resolution, but it would be nice if resolution was renamed to something that everyone can understand. My suggestion would be "quality".

I was aiming for a shorter name to not have it cut off and reso sounded catchy lol
that said yeh quality is definitely more descriptive

> Spread should not have a button to toggle a slider between 0.01 and the max

It started as a button, later turned to a slider, then I increased lower limit to 0.01 because having it at 0 bugged something— that said it makes a lot more sense to have it be just a slider correct yeh and I'll just turn it off if it's 0.

> Also, it's called "scatter" in one place and "spread" in another

You're right! Idk how I missed that ahh few weeks back I called it scatter later in the code I just decided to call it spread when I made the slider 😓 

> We should consider removing some of the preset shapes and/or putting them in a dropdown.

A dropdown is possible but an easier option I was actually considering is horizontal scroll 🤔 

> When editing a brush, the pressure, relative size, and mode buttons should not be there. They are accessible when outside of the brush editor because they can change at any time without "altering the fundamental brush". Therefore, the brush editor should not show these, as it's only for the "important" settings.

Good point here this I can do
Also, in regards to all the points with the trash can button… might be a good idea to have a mini "x" that appears when hovering over the brush presets/ shape options… 

> 3 rotation options: set offset angle (default is zero), follow mouse (with offset angle), and randomize angle.

Agree with you there 

> When dragging the sliders, the preview flickers quite a bit, which is distracting
I noticed… kinda a given since we're updating an image element and re-rendering. I've thought of that and a way to stop it— was just too lazy to work on it but should be doable

> When using a big brush size, the brush path before releasing the mouse is pretty low resolution. Is it possible to increase the resolution so that it looks nicer?
I'll say "no" to that _for now…_ but will see what i can do there

> We should have preset brushes that cannot be significantly modified or deleted
I can agree with that that said we'd need to have some indicator there or just Idk— that said an "adobe brush" has been demanded and would be great to have as a preset option

> There should be a cancel button to exit the brush config without saving it.
Perhaps adding a "<-" button up there next to the name parameter, then replacing the save brush button with a "reset settings" 🤔

Alani25:
Note to self: Allow drag and dropping `.cbrush` files to import brush assets

BaronAWC:
Few more things I think we should change:
- [x] The default brush should be plain
- [x] There is an old bit of code that adds 1 to the brush size, which we should remove
- [ ] Brush scatter internally has some magic numbers, being 1.4 and 2.8 somewhere. After i figure out what they're there for, we should consider removing them, rounding them to nicer numbers, etc.
- [x] Relative brush size should be off by default
- [ ] The sparkle behaves poorly if it's too small. We should consider making it a little less thin on the edges
- [ ] The highest quality should(?) be higher than it currently is
- [ ] The preview should show a slightly more smaller, zoomed-out version. It's hard to see the true effect of settings when it's too zoomed in
- [ ] The preview doesn't look very accurate, and it seems to have a LOT of magic numbers. After I give it a closer look I'd like to reduce the magic numbers as much as possible
- [x] The brush creator/editor should not have a setting for size and lead
- [ ] The scatter brush shape should be removed, as it's basically a redundant version of the circle with built-in spread. We could probably bring back a different shape
- [ ] I think the UI still has room for improvement but I'm not 100% sure what we should do right now... I'll do some research and talk over button placement when the time comes
    - [ ] A brush should be selected the first time you enable the brush tool
    - [ ] The inspector title should(?) say "Brush Tool" instead of the brush's name, to match how other titles work (e.g. clips with names)
    - [ ] Downloading a brush (and maybe deleting it) should not be hidden behind editing it
    - [ ] The fixed angle mode should make the brush's icon set at the fixed angle, in the brush selector and in the canvas (e.g. a 45-degree offset square brush should show as a diamond)
    - [ ] The path-angle mode should make the brush's canvas preview rotate to follow the brush

I'll do a couple of these really quick tonight. I'll see what else I can do over this upcoming week.

src/Editor/Panels/Inspector/Inspector.jsx
```diff
@@ -36,17 +36,68 @@ import InspectorSoundPreview from './InspectorPreview/InspectorPreviewTypes/Insp
...
+
 import { Console, Hook, Unhook } from 'console-feed';
 // import { useEffect, useState } from 'react';
 
 window.EditorGradientColorSwapState = false;
 
+// Default brushes — structured like a parsed .cbrush JSON for easy migration later.
+// Each field matches the keys stored/loaded by saveBrush / applyBrush.
+// FIXME: these "default" values are in 3 places (search for this same comment)
+const DEFAULT_BRUSHES = [
```
BaronAWC, line 54: I feel like this default brush doesn't have to be in Inspector.jsx, it makes more sense either in Brush.js or possibly somewhere else

src/Editor/Panels/Inspector/Inspector.jsx
```diff
@@ -36,17 +36,68 @@ import InspectorSoundPreview from './InspectorPreview/InspectorPreviewTypes/Insp
...
+    brushSpacing: 0.2,
+    brushScatterAmount: 0,
+    brushRotationMode: 'fixed',
+    brushRotationOffset: 0,
+    brushStabilizerWeight: 20,
+    fillColorRgba: '#000000',
+  },
+];
+
+const BRUSH_SHAPES = [
```
BaronAWC, line 69: I would imagine this is redundant, since the brush shapes exist somewhere else. When creating a new brush shape, the editor also knows how to create the preview, so I'd imagine it's doable for these default shapes too.

src/Editor/Panels/Inspector/Inspector.jsx
```diff
@@ -36,17 +36,68 @@ import InspectorSoundPreview from './InspectorPreview/InspectorPreviewTypes/Insp
...
+    svg: <rect x="2" y="2" width="24" height="24"/> },
+  { id: 'sparkle',    name: 'Sparkle',
+    svg: <polygon points="14,1 15.4,12.6 27,14 15.4,15.4 14,27 12.6,15.4 1,14 12.6,12.6"/> },
+  { id: 'leaf',       name: 'Leaf',
+    svg: <path d="M14,2 Q26,14 14,26 Q2,14 14,2 Z"/> },
+  { id: 'scatter',    name: 'Scatter',
+    svg: <><circle cx="14" cy="14" r="4"/><circle cx="7" cy="8" r="3"/><circle cx="21" cy="8" r="2.5"/><circle cx="7" cy="20" r="3"/><circle cx="21" cy="20" r="2.5"/></> },
+];
+
+const ROTATION_MODE_OPTIONS = [
```
BaronAWC, line 82: These only seem to be used once, so I wonder if it makes more sense to hardcode

src/Editor/Panels/Inspector/Inspector.jsx
```diff
@@ -134,6 +212,337 @@ class Inspector extends Component {
...
+      // Tangent angle of x(t)=margin+t*pathW, y(t)=H/2+sin(t*pi*2.5)*amplitude
+      const dxdt = pathW;
+      const dydt = amplitude * Math.cos(t * Math.PI * 2.5) * (Math.PI * 2.5);
+      return Math.atan2(dydt, dxdt) + offsetRad;
+    }
+    return offsetRad; // 'fixed'
+  }
+
+  // BRUSH PREVIEW— we're literally drawing a brush stroke here -H.A.
+  drawBrushPreview = () => {
```
BaronAWC, line 261: Could the brush preview go to a separate file? It's not really suitable in the inspector file

src/Editor/Panels/Inspector/Inspector.jsx
```diff
@@ -930,6 +1339,696 @@ class Inspector extends Component {
...
+  startNewBrush = () => {
+    this._captureEditSnapshot();
+    this.setState({ selectedBrushIndex: null, isEditingBrush: true, showNewBrushMenu: false });
+  }
+
+  importBrush = () => {
+    if (this.brushFileInputRef.current) this.brushFileInputRef.current.click();
+  }
+
+  handleBrushFileImport = (e) => {
```
BaronAWC, line 1367: Import/export could probably go into a different file - doesn't make sense in inspector

src/Editor/Panels/Toolbox/ToolSettings/ToolSettings.jsx
```diff
@@ -70,10 +70,6 @@ class ToolSettings extends Component {
     return (
       <div className='settings-input-container'>
         {this.renderBrushSize()}
```
BaronAWC, line 72: we may not even need brush size here, since it's in the inspector now

src/Editor/Panels/Toolbox/ToolSettings/ToolSettings.jsx
```diff
@@ -234,6 +230,19 @@ class ToolSettings extends Component {
     )
   }
 
+  renderBrushResolution = () => {
```
BaronAWC, line 233: this is unused?

engine/src/tools/Brush.js
```diff
@@ -349,17 +374,205 @@ Wick.Tools.Brush = class extends Wick.Tool {
...
+                }
+                ctx.restore();
+                var hsCustom = Math.round(S / 2);
+                return 'url(' + canvas.toDataURL() + ') ' + hsCustom + ' ' + hsCustom + ', default';
+            }
+            // Custom shape data not loaded for some stupid reason? fall back to circle below
+        }
+
+        // Crescent needs compositing — handle separately
+        if (shape === 'crescent') {
```
BaronAWC, line 433: Can the brush shapes go into a separate file?

BaronAWC (changes requested):
Code-related comments:

Main comments here are about moving code out of Inspector.jsx and into either Brush.js or a new file(s).

Additionally, in the code itself, we should standardize terms like spread/scatter, resolution/quality, etc.

BaronAWC:
I quickly fixed the default brush settings in one final place, which should make it consistent with the other 3 places. (Speaking of which, we should try to only have it in one place)

Also, I'm not sure how to reproduce this, but sometimes the fill color of the brush preview will disappear entirely, leaving just the outline. Has anyone else encountered this?

StickmanRed:
> I quickly fixed the default brush settings in one final place, which should make it consistent with the other 3 places. (Speaking of which, we should try to only have it in one place)

I think we should try to move all the brush shape-drawing logic into an engine file, maybe `tools/BrushUtils.js`. The logic is replicated several times, just with different color, position, or rotation, but relative to those factors, the shape-drawing commands should be the same. This could be a function `Wick.Brush.Utils.drawBrushShape(ctx, size, shape, args)` that can be used for the [cursor image](https://github.com/Candlestickers/Candlestick/blob/416595d9e4636ce1546024bc92da26c1a28e0a5e/engine/src/tools/Brush.js#L387), [brush tip](https://github.com/Candlestickers/Candlestick/blob/416595d9e4636ce1546024bc92da26c1a28e0a5e/engine/src/tools/Brush.js#L786), and [brush preview](https://github.com/Candlestickers/Candlestick/blob/416595d9e4636ce1546024bc92da26c1a28e0a5e/src/Editor/Panels/Inspector/Inspector.jsx#L437), saving duplicated logic and ensuring a consistent brush shape.

PR #174, opened 2026-07-15
node24 Package Clean

Alani25:
# >20 Modified Packages

Unused/ removable packages:
- `url-parse` causes the baseURL depreciation warning, not really used anywhere just credited, and has known CVEs, it's too basic of a library if anything better to just use modern `url` constructor
- `react-measure` zero imports
- `react-tiny-popover` zero imports and we use `react-popover` instead of this so it's useless
- `dotenv` zero imports, here environmental variables are handled by CRA so this is useless
- `rc-slider` zero imports
- `react-tabs` zero imports
- `react-dropzone` zero imports
- `react-color`
- `react-aria-menubutton`

There were also 4 libraries I could just upgrade safely bc they're minor upgrades 
- `ace-builds`
- `react-spinners`
- `react-dropdown`
- `base64-arraybuffer`
- `underscore` - replaced with a basic 8-line throttle function in `Editor.jsx`
- `query-string` - replaced with native URLSearchParams in `EditorCore.jsx`

Upgraded dependencies
- `react-device-detect` **v1 -> 2**
- `react-dnd` **v11 -> 16** (HOC to hooks migration)
- `react-tooltip` **v4 -> 5**
- `react-reflex` **v3 -> 4**
- `react-select` **v2 -> 5**
- `react-ace` **v6 -> 14**
- `toast` **v4 -> 9** (not highest)

Replaced dependencies
- `react-popover` for reactstrap's `Popover` (see number value slider)
image
- `react-sizeme` replaced with `react-resize-detector`, an unused dependency but has been maintained longer

# ~12 Outdated Packages

Run `npm outdated` in terminal to view remaining outdated dependencies
image

Most of the remaining 12 dependencies either moved on from CRA to Vite (CRA is currently unmaintained), or require a higher react version (we're currently on react 16)
<sup> Note: vulnerabilities ≠ outdated packages</sup>

The only reason why I'm not going through the trouble of updating react further or CRA is because all the remaining changes are either only effective in development (building and what not, ex: gulp), or the UI, and unrelated to software performance. We don't have much issues with the editor UI, and such fixes would be barely noticeable. In comparison, there's far more risk when updating. 
On another note, CRA is unmaintained has been replaced with vite, however a lot of our current dependencies we have aren't compatible with vite. We can go through the trouble of upgrading them all, but the only benefit to this would be building the app would go from 15 seconds to 5 seconds.
This is why, after having updated what I can and "cleaning" out what we do not need/ use, I think we're at a good point in terms of dependencies.

# Performance and Memory
performance wise— found an issue where we'd be constantly unmounting-remounting the outliner based on a `Math.random()`, changed it to happen based on timeline edit. 
Another problem where loading MP4s (from importer) didn't actually close the audio context function (on me), so if the user opened a decent number of videos without reloading, the browser would reach the limit and the next vid wouldn't have an audio. Not that it'd be a noticeable issue since you usually open one file and work there.
That said, all these points combined seem to have an effect. Images below from inspector go to prove this.

Left images: Old
Right images: New

image
image 

image
image

I've also taken multiple snapshots of the editor memory, and it is important to note that the old editor saved copies of all projects opened within the same session. For instance, if you open a big project, then click "new" to start a new one or open a different file because you opened the wrong one, the older project is still saved in memory.
I tested this by flipping through different projects in the same session. As you can see, on the old editor multiple wick projects are still in memory (x13 in this testing case).
image
The thing is even when clicking undo, you do not go to the previous project, so that data is really just taking up space. In this PR, we toss the older projects out when we open a new project. The memory numbers may still show the new and old editor taking the same amount of memory, but that'd be because after it's been "garbage tossed," chrome allocates that memory in case it's used again, but most importantly if the device needs it it'll take it away.

# Extra notes
After updating the toast library new icons were added. I plan on removing these.
image image
image image

While I mentioned above the baseURL was not used, I later realized where it should've been used and the fix was applied to the development branch. That should be easily fixed during the merge process.

Replacing the abandoned `react-popover` for reactstrap's `Popover` did slightly "break" the popover (left image), but I fixed the issue in this PR (right image). The difference between this and the current editor popover is barely noticable.
Feel free to test.
image image

Replacing `react-sizeme` (unmaintained) with `react-resize-detector` actually improved the dragging tiles— in the past if you drag the panels quick you can catch a glimpse of a gray background box, which is now not there. This gives you a good idea of how unnoticeable upgrades to react and UI based dependencies are.
[Discord Quick Video](https://discord.com/channels/1406876763920797717/1523695010652881057/1523695358843027539)

Amount of vulnerabilities has gone down to 40, and they're not critical. 
The paperJS is an unreported and more critical XSS vulnerability that'll be handled on a separate branch than this one.


InternetAstronaut:
Quick note, we should also look over the colors to see if they match WCAG 4 standards. It's easy to notice that some parts of the editor (many of those we edited, since we don't have the original design files) are very hard to read.

Alani25:
> Quick note, we should also look over the colors to see if they match WCAG 4 standards. It's easy to notice that some parts of the editor (many of those we edited, since we don't have the original design files) are very hard to read.

This could tie into another branch more focused on the UI— that said I can also see how this relates to some of what I did in this branch. For example the toast notifications, the icon and colors are customizable there. 
We should take the main panel and scss defined colors and provide the setting options for users to set at their own discretion, with a collection of preset themes

Though a bit old, here's a map from my notes you'd be able to use to find where most of the colors in the code are. Of course once again, this can be carried into a new branch and we can work on it there
image

PR #176, opened 2026-07-16
fix: Increase width of opaque stroke color display

StickmanRed:
I *just* noticed this


InternetAstronaut (approved):
LGTM

BaronAWC (approved):
I suppose one thing we can consider is to turn the 4-pixel thickness of the color picker into a variable. That way, these magic numbers make a bit more sense, and if for whatever reason we change the line width, we can just do it once.

StickmanRed:
Good point, I'll go ahead and move it to a variable

StickmanRed:
Untested but should work

PR #179, opened 2026-07-27
Migrate CRA to Vite

hobbsythe6th:
Migrated CRA to Vite and removed a lot of scss deprecation warnings that Vite likes throwing. Also, for some reason our jsx has require('classnames') in basically every file, which I fixed. AND, since the jsconfig.json data is now in vite.config.js, the jsconfig has been deleted with no ill effects!


InternetAstronaut:
Vite migration is something we really need, the extra 5 seconds may look stale until you notice this means we can stop doing things like almost patching packages that CRA requires.

StickmanRed:
I think we should split this into two PRs: one with just the `@import` to `@use` and `require()` to `import` changes, and then one with the CRA to Vite changes. This would make it easier to review and focus on the Vite changes

hobbsythe6th:
I know, but I'm not experienced enough with Git to move a singular commit to another branch/delete singular commit. Also, even though CRA throws the scss stuff, I didn't find it till the Vite tests so it ended up in the Vite mess. Vite's much stricter about ESM/CJS so the require() thing was also a new mid-PR discovery. I'll try and put it in another PR tho

StickmanRed:
Sounds good! Though let me know if you'd want any help with the process

hobbsythe6th:
I think I reverted the commits, but it makes this PR untestable sadly. but I can't believe how much easier to read the diff is!

src/Editor/Modals/BuiltinLibrary/BuiltinLibrary.jsx
```diff
@@ -40,7 +40,7 @@ class BuiltinLibrary extends Component {
     if (window.__TAURI__) {
       return 'builtinlibrary/'; // resources folder will be flattened
     }
-    return process.env.PUBLIC_URL + '/builtinlibrary/';
+    return import.meta.env.VITE_PUBLIC_URL + 'builtinlibrary/';
```
StickmanRed, line 43: I'm not too sure how `process.env.PUBLIC_URL` works and what its value is in production builds... this is just a reminder for myself to double-check these parts and make sure they will work in the candlestick page

hobbsythe6th: as far as I know process.env is an object CRA injected into the build that just has everything in the .env file.  import.meta.env is Vite's version of that

PR #180, opened 2026-07-28
Add moveable pivot to selection widget and clips

StickmanRed:
### Features
- Drag pivot handle to reposition
- When a single clip is selected, clip saves pivot offset
- Pivot handle locks when the selection box is too small (when both width and height are smaller than 4x the pivot size)
- Alt to rotate from opposite corner

Also the selection code was significantly cleaned up. I had previously contained all the changes to the selection tool in a `this.mod` object... no more of that. Some of the logic was simplified as well.

### DIFF
--- a/engine/src/base/Clip.js
+++ b/engine/src/base/Clip.js
@@ -53,6 +53,7 @@ Wick.Clip = class extends Wick.Tickable {
         this._isSynced = false;
 
         this._transformation = args.transformation || new Wick.Transformation();
+        this._pivot = args.pivot || [0,0];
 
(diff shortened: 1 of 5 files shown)


Alani25 (approved):
Tested, works as expected and offset pivot information saves in file as well. 
Code is well structured.

My only suggestion is to add in a new cursor style for when the cursor is hovering over the pivot point— that or adjust the color of the pivot point to indicate it's being hovered over and draggable.
That's more of an additional UI/ UX tweak but with it aside this PR should be ready to merge

PR #184, opened 2026-08-07
Update engine to use (some) packaged libraries instead of files

hobbsythe6th:
This included migrating Gulp to Rolldown, as Gulp does not have a resolver. Also, since Rolldown comes as a Vite dependency, I also merged the Vite migration into this. There are probably a lot of bugs, but all the ones I found with the new setup were patched (except for one: on opening the page the timeline is only partially rendered. But it re-renders after you close the welcome modal so I don't think that it's very significant.)
I also modified the index.html to load paper and jQuery as separate scripts, as I didn't have the patience to add any integration.
Since this was really hard, I started out by using Webpack, so that's why the first few commits have it. The Rolldown switch came later.
BTW, I didn't make Croquis a package because I didn't have the patience to figure the package structure out, so I just stuck with the local file. Also, some libraries like convertRange are from stackoverflow so I left those alone.

Read the standalone diff at https://github.com/hobbsythe6th/Candlestick/pull/1


hobbsythe6th:
new bugs found, working on it currently

src/index.jsx
```diff
@@ -32,9 +32,11 @@ initAndroidPlatform();
...
 
+window.CandlestickVersion = '1.0.3';
+
 ReactDOM.render(<Editor />, document.getElementById('root'));
 
 // If you want your app to work offline and load faster, you can change
 // unregister() to register() below. Note this comes with some pitfalls.
 // Learn more about service workers: http://bit.ly/CRA-PWA
-serviceWorker.unregister();
+serviceWorker.unregister();
```
InternetAstronaut, line 42: putting `serviceWorker.register();` will speed up the app.

The _'pitfall'_ is that we won't have support for iOS 11.2 or earlier, which isn't a pitfall anymore.

hobbsythe6th:
fixed and added a lot of checks! may need testing tho

InternetAstronaut (commented):
forgot to submit it...

src/index.jsx
hobbsythe6th: I have no idea what this does so I'm just leaving it as is

InternetAstronaut (approved):

engine/package.json
```diff
@@ -7,8 +7,29 @@
...
+    "howler": "^2.1.1",
+    "hull": "AndriiHeonia/hull",
+    "invert-color": "^2.0.0",
+    "is-var-name": "^2.0.0",
+    "jquery": "^3.3.1",
+    "jquery-mousewheel": "^3.1.13",
+    "jszip": "^3.1.5",
+    "lerp": "^1.0.3",
+    "localforage": "1.7.3",
+    "paper": "^0.12.4",
```
InternetAstronaut, line 24: I recommend biting the bullet and changing to Scratch's Paper, all features removed in their fork (built-in tweening and PaperScript) aren't used in Candlestick.

engine/package.json
```diff
@@ -7,8 +7,29 @@
...
+  "dependencies": {
+    "@tweenjs/tween.js": "^17.4.0",
+    "base64-arraybuffer": "^0.2.0",
+    "esprima": "^4.0.1",
+    "floodfill": "^2.1.1",
+    "howler": "^2.1.1",
+    "hull": "AndriiHeonia/hull",
+    "invert-color": "^2.0.0",
+    "is-var-name": "^2.0.0",
+    "jquery": "^3.3.1",
```
InternetAstronaut, line 19: Why do we need jQuery `3.3.1` specifically, we should either use `4.x` or latest `3.x`.

engine/package.json
```diff
@@ -7,8 +7,29 @@
...
+    "jquery-mousewheel": "^3.1.13",
```
InternetAstronaut, line 20: Is this ever used? Should we use it? Can we remove quickly?

engine/package.json
```diff
@@ -7,8 +7,29 @@
     "lib": "lib",
     "test": "tests"
   },
+  "dependencies": {
+    "@tweenjs/tween.js": "^17.4.0",
```
InternetAstronaut, line 11: `Tween.js`'s last update was 2 years ago (2024). Not only that, it's also blocking.

I know we're not trying to minimize package use, but this may be impacting performance.

We should also change from `^17.4.0` to `25.0.0` (latest version)

engine/package.json
```diff
@@ -7,8 +7,29 @@
...
+    "base64-arraybuffer": "^0.2.0",
+    "esprima": "^4.0.1",
+    "floodfill": "^2.1.1",
+    "howler": "^2.1.1",
+    "hull": "AndriiHeonia/hull",
```
InternetAstronaut, line 16: Deprecated package, replacement needed.

engine/package.json
```diff
@@ -7,8 +7,29 @@
     "lib": "lib",
     "test": "tests"
   },
+  "dependencies": {
+    "@tweenjs/tween.js": "^17.4.0",
+    "base64-arraybuffer": "^0.2.0",
+    "esprima": "^4.0.1",
+    "floodfill": "^2.1.1",
```
InternetAstronaut, line 14: Old package, last updated 2019. Slow, and pollutes both global window context and `CanvasRenderingContext2D` prototype.

engine/package.json
```diff
@@ -7,8 +7,29 @@
...
+    "jquery": "^3.3.1",
+    "jquery-mousewheel": "^3.1.13",
+    "jszip": "^3.1.5",
+    "lerp": "^1.0.3",
+    "localforage": "1.7.3",
+    "paper": "^0.12.4",
+    "platform": "^1.3.5",
+    "pressure": "^2.1.2",
+    "quadtree-lib": "^1.0.9",
+    "uuid": "^14.0.1"
```
InternetAstronaut, line 28: What about the faster `nanoid` library? It's as safe, faster, and nobody will notice it being changed anyway.

InternetAstronaut: Wick also doesn't check on UUIDs.

InternetAstronaut (changes requested):
In general, it's just a simple refactor.

engine/package.json
hobbsythe6th: we should do this in another PR imo, this is just a migration, not really an 'update engine packages' PR

engine/package.json
hobbsythe6th: jquery is in two or so places i think, is-var-name is used in Wick.Base and invert-color is in the brush. don't remember about jquery-mousewheel but i had to include it in the index.html to prevent zooming out with mousewheel breaking the editor

engine/package.json
hobbsythe6th: im just mirroring the versions in the old lib folder, so there might be breaking changes. again, probably for another subsequent PR

engine/package.json
hobbsythe6th: yet again: different PR for a package bump

engine/package.json
hobbsythe6th: idk, but the uuid swap is a great idea imo: for another PR.

engine/package.json
hobbsythe6th: im assuming you mean floodfill, and I'm not bumping any deps here, just making it possible.

engine/package.json
hobbsythe6th: im just using the one that we currently use

PR #185, opened 2026-08-08
Changed PWA manifest to correctly name Candlestick

InternetAstronaut:
The `manifest.json` file (the one that makes it possible to _'save'_ candlestick as an app on your homescreen) still **incorrectly called the app 'Wick Editor'**.


Alani25:
LGTM
Nice catch thanks man

PR #191, opened 2026-08-16
Video importer into asset library + extra video import formats

Alani25:
Users can now import MP4s through the asset library and do not need to open MP4s as a completely separate project, allowing them to add videos as assets into an existing project.

I believe the original MP4 import code I was able to get through stack overflow, but it wasn't optimized enough for our use case. So, while working on this I also added multiple workers for the MP4 importer to speed up the process, and also had the audio extraction work in parallel as well… no idea if this would improve the audio quality but the speed difference is noticable.

Importing a test video of length 1 minute 42.5 seconds at an framerate of 12 FPS resulting in up to ~1230 frames, using a stopwatch to record the import time we get:
- Old MP4 importer: 3 minutes 17 seconds
- With 4 workers: 1 minute 49 seconds (1.8x faster)
<sup>^ I mean zero surprises here it only makes sense</sup>

Theoretically we can go for more workers on better devices, specifically with the tauri builds, but 4 workers seems to be the right bar for all devices so far (especially Linux). This may come at a later time.

One last improvement— "MP4" importer now accepts other video formats (mov, webm, mkv, avi, 3gp, ogv, m4v, wmv). 
Thanks to the way the current MP4 importer system is set up this was just a matter of allowing the user to select another format. Haven't tested on all formats, but mov and MP4 work (mov is the same format used by apple devices by default, making video import on iphones a lot easier as well).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,6 @@
         "file-saver": "^2.0.5",
         "gifuct-js": "^2.1.2",
         "localforage": "^1.10.0",
-        "mp4box": "^2.4.1",
         "pdfjs-dist": "^6.1.200",
         "react": "^16.13.1",
         "react-ace": "^14.0.1",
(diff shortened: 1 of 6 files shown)


InternetAstronaut:
I suggest using a **different approach on Tauri/Desktop**, called **OFFLOAD IT USING REAL FFMPEG**.

PR #193, opened 2026-08-19
Wick.Base extensibility

hobbsythe6th:
I was inspired by paperjs' inject() function so I added this one. It allows you to define a custom function for any instance of any class derived from Wick.Base . You can't add a custom API into EVERY instance yet, but I think this is a good starting point for the engine to become more customisable.
Also, if you want to you <i>can</i> technically just use it to define a new property but it's a bit too much hassle.

Branched off of the engine update branch, which in turn is branched off of the vite migration. I'm sorry, but I just can't live with CRA/Gulp anymore.

Edit: parameters are now passed as arrays, and this has been added to the code reference


InternetAstronaut:
You just implemented the way **extending classes** was done **before ES6**. (**before 2015**, depending on the dev they switched in **2016 or 2017** to more modern practices)

The problem with this method is not only is prototypal reflection slow in V8 (they use some optimizations which make it slower), prototypes affect **ALL INSTANCES**.

The only way to track and localize the extensions made is creating a brand new class which extends Wick.Base.

This will also affect editor code, since it only resets when the prototype is re-defined (e.g.: reloading the page).

It's better to just do `(Wick.Clip /*instance*/).x = y` than this.

hobbsythe6th:
I'm new to JS so I understood nothing :/
Anyway I'll close the PR if this is unnecessary.

PR #195, opened 2026-08-19
Asset folders

hobbsythe6th:
Added asset folders.
Assets and folders can both be dragged and dropped onto other folders to move them in, and can be dropped on the 'back' button to move them out.

Branched this off the Vite migration, read the diff at https://github.com/hobbsythe6th/Candlestick/pull/6


src/Editor/Panels/AssetLibrary/AssetLibrary.jsx
```diff
@@ -19,15 +19,46 @@
...
 
+/**
+ * The header shown above the asset grid while inside a folder. Doubles as
+ * a drop target: dropping an asset or folder here moves it out to the
+ * parent level, since there's otherwise no way to drag something out of
+ * the folder you're currently viewing.
+ */
+function FolderHeader(props) {
+  const [{ isOver }, drop] = useDrop(() => ({
+    accept: ACCEPT_TYPES,
```
InternetAstronaut, line 41: I don't do the UI thing, but if `react-drop` accepts functions, it would be easier to just do:
```js
i => (i instanceof String || typeof i === "string") ? i.endsWith('Asset') : false
```

InternetAstronaut (approved):
LGTM

src/Editor/Panels/AssetLibrary/AssetLibrary.jsx
hobbsythe6th: apparently you can't do that sadly

PR #198, opened 2026-08-21
fix: Line tool diagonal snapping and angle

### LINKED ISSUE #28: Straight Lines with Line Tool
Similar to how you can hold <kbd>shift</kbd> while moving objects to move them in a straight path, you should also be able to hold <kbd>shift</kbd> while drawing with the line tool to create a straight line.

StickmanRed:
### Description
Fixes two issues in the Line tool: line snapping not working when drawing a diagonal line, and angle to draw diagonal lines being slightly off.

### Testing
1. Open the Line tool and draw a line on the canvas.
2. Hold `Shift` to draw another line diagonally, and snap to a vertex on the first line. When snapping, the line's endpoint should not move with the mouse.
3. Move the endpoint to a `22.5` degree angle, halfway between diagonal and horizontal/vertical. The line should switch between diagonal and horizontal/vertical around this halfway angle.

https://github.com/user-attachments/assets/7002dfba-44bc-419a-869a-89aae0bef8b0

### DIFF
--- a/engine/src/tools/Line.js
+++ b/engine/src/tools/Line.js
@@ -160,15 +160,16 @@ Wick.Tools.Line = class extends Wick.Tool {
 			const dy = Math.abs(this.startPoint.y - e.point.y);
 			const dx = Math.abs(this.startPoint.x - e.point.x);
 
-			// diagnol
-			if (dy && dx && (dy / dx) > 0.5 && (dx / dy) > 0.5) {
+			// diagnol (tan(22.5) = 0.41421356237309503)
+			if (dy && dx && (dy / dx) > 0.41421356237309503 && (dx / dy) > 0.41421356237309503) {
 				const diff = {
 					x: this.endPoint.x - this.startPoint.x,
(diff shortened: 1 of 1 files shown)


engine/src/tools/Line.js
```diff
@@ -160,15 +160,16 @@ Wick.Tools.Line = class extends Wick.Tool {
 			const dy = Math.abs(this.startPoint.y - e.point.y);
 			const dx = Math.abs(this.startPoint.x - e.point.x);
 
-			// diagnol
-			if (dy && dx && (dy / dx) > 0.5 && (dx / dy) > 0.5) {
+			// diagnol (tan(22.5) = 0.41421356237309503)
+			if (dy && dx && (dy / dx) > 0.41421356237309503 && (dx / dy) > 0.41421356237309503) {
```
Alani25, line 164: This makes a lotta sense, when I originally wrote this code I noticed this issue but didn't give it much thought then.

Alani25:
Code and functionality looks good to me

PR #199, opened 2026-08-22
fix: Text box doesn't focus on first click

### LINKED ISSUE #96: [BUG] when creating a new text box, it is not focused
**Describe the bug**
when creating a new text box, you have to click on the text box to be able to type, delete, select, etc.

**To reproduce**
1. create a text box
2. try typing or using cmd+A or ctrl+A - the typing doesn't work
3. click on the text box again - now the typing works

**Expected behavior**
text box is immediately editable without manually focusing it

**Screenshots**

**Editor version**
only candlestick, but also only the browser version. though, the desktop one has... its own problems. 

**Device**
 - macbook air m4, chromium browser
 - windows 10, chrome

i tried this on a very old branch and the issue was gone, which means we messed something up (which, i guess is a good thing?)

this is the commit where the issue first arises: https://github.com/Candlestickers/Candlestick/pull/66/commits/879f180ad047170fb666b5bfea263f3a305b3b50
- it is somewhere in View.Project.js, where the bulk of the mobile gesture code lives. the changes in Tool.js and the gestureActive() function in View.Project.js do not seem to be part of the problem.

**Additional context**
https://discord.com/channels/1406876763920797717/1406888260591947867/1443037923518844978

StickmanRed:
### Description
The gesture code in `View.Project.js` uses pointer capture to funnel mouse events onto the `<canvas>` element, which triggers [an `onclick` event](https://github.com/Candlestickers/Candlestick/blob/4266827a3c460fee7d9eb1120578b7c83c27ce52/engine/src/view/View.Project.js#L585) that defocuses the Text tool's text box.

This PR implements a hacky fix to avoid breaking the gesture code. `tools/Text.js` wraps the canvas `onclick` handler in a function that refocuses the text box then sets the `onclick` handler to the original handler.

Closes #96

### Testing
1. Switch to the Text tool.
2. Press left mouse button and release in the same location. The cursor should be focused on the text box.
3. Click anywhere else on the canvas to deselect the text.
4. Click back on the text to select it. The text box should be focused again.
5. Deselect the text.
6. Press left mouse button on the text, but drag the cursor outside the text box and release. The text should be focused again. **This behavior is different from Wick Editor 1.19.3**, where the text box would be deselected.

Personally, I like having the text focused in this case. ~Not a bug, but a feature lol~

### DIFF
--- a/engine/src/tools/Text.js
+++ b/engine/src/tools/Text.js
@@ -76,11 +76,18 @@ Wick.Tools.Text = class extends Wick.Tool {
             return;
         }
 
+        let oldTargetOnClick = e.event.target.onclick;
+        const targetOnClick = targetEvent => {
+            oldTargetOnClick && oldTargetOnClick(targetEvent);
+            e.event.target.onclick = oldTargetOnClick;
+            this.editingText.focus();
+        }
         if (this.editingText) {
             this.finishEditingText();
         } else if(this.hoveredOverText) {
             this.editingText = this.hoveredOverText;
             e.item.edit(this.project.view.paper);
+            e.event.target.onclick = targetOnClick;
         } else {
             var text = new this.paper.PointText(e.point);
             text.justification = 'left';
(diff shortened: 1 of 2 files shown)


Alani25:
YES, massively improving the UX with this one
I've tested the pr locally and all seems to work as expected :)
Code also looks good and pretty straight forward.
Appreciate the work you've put in here.

PR #200, opened 2026-08-22
fix: Paths edited by Path Cursor can sometimes retain detailed handles

StickmanRed:
### Description
If you use the Path Cursor's detailed editing on a path, on occasion, the path handles can somehow get stuck to the path. In this PR, each path checks if the project (if it exists) is using the Path Cursor tool, and if it ain't, the path quietly discards the selection handles and moves on.

### Testing
1. Create some paths.
2. Activate the Path Cursor.
3. Double-click on a path to enter detailed editing mode.
4. Move a vertex.
5. Switch to another tool. The detailed editing handles should be removed.
6. Undo once. The handles should still be removed. In previous versions the handles still appear.

### DIFF
--- a/engine/src/base/Project.js
+++ b/engine/src/base/Project.js
@@ -1171,6 +1171,8 @@ orderDynamicFrames() {
             });
         });
 
+        // Switch to cursor tool so we can interact with the selection
+        this.activeTool = 'cursor';
         this.selection.selectMultipleObjects(objectsToAdd);
     }
     
--- a/engine/src/tools/PathCursor.js
+++ b/engine/src/tools/PathCursor.js
@@ -275,7 +275,7 @@ Wick.Tools.PathCursor = class extends Wick.Tool {
             if (targetItem && targetItem.parent.className === 'CompoundPath') {
                 targetItem = targetItem.parent;
             }
-            if (this._getWickUUID(targetItem) !== this._getWickUUID(this.detailedEditing)) {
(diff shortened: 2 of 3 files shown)


Alani25:
Looks good— only came by one issue while testing—
If you use <kbd>ctrl</kbd> + <kbd>z</kbd> _while using_ the path cursor tool (double click to enter path cursor edit mode of shape, move little, then undo), then re-double clicking with the path cursor to get back into the path cursor edit mode doesn't seem to work.

Tested this on the older editor version and can confirm the same bug is there as well. 
If we can cover it in this same PR would be great.

And one more quick thing that might be worth adding— when using the <kbd>ctrl</kbd> + <kbd>a</kbd> shortcut to select all objects in the frame (while using the path cursor, text tool, or any other tool asides the cursor tool) we should allow it to automatically switch to the cursor tool, otherwise everything gets selected and you can't move it around because you're still on the current tool.

InternetAstronaut:
This is because the detailed path editing mode is ported from Legacy Wick. I believe that we should re-do the Path Cursor.

StickmanRed:
I got confused finding a fix for the first problem, because Wick creates new path objects every render... kinda wish we had more control over the `paper.js` objects

InternetAstronaut:
> kinda wish we had more control over the `paper.js` objects

That's the problem with extremely abstract libraries. You can't control everything.

PR #201, opened 2026-08-23
Frame size slider

Alani25:
Added setting option for frame size to allow users to minimize the frame size smaller than the current, smallest limit.
Smaller frame size options have been requested multiple times on Discord by ~2 users. 

The setting option is a slider. 
This increases customizability to make the timeline feel closer to what may be more comfortable working with.

### DIFF
--- a/engine/src/gui/Frame.js
+++ b/engine/src/gui/Frame.js
@@ -89,11 +89,23 @@ Wick.GUIElement.Frame = class extends Wick.GUIElement {
             this.cursor = 'grab';
         }
 
+        // Dot scale shared by script indicator and content dot
+        var _dcw = this.gridCellWidth;
+        var _dG = Wick.GUIElement;
+        var _dotScale;
+        if(_dcw <= _dG.GRID_SMALL_CELL_WIDTH)
+            _dotScale = 0.5 + 0.25 * (_dcw / _dG.GRID_SMALL_CELL_WIDTH);
+        else if(_dcw <= _dG.GRID_NORMAL_CELL_WIDTH)
+            _dotScale = 0.75 + 0.25 * ((_dcw - _dG.GRID_SMALL_CELL_WIDTH) / (_dG.GRID_NORMAL_CELL_WIDTH - _dG.GRID_SMALL_CELL_WIDTH));
+        else
+            _dotScale = 1.0 + 0.25 * ((_dcw - _dG.GRID_NORMAL_CELL_WIDTH) / (_dG.GRID_LARGE_CELL_WIDTH - _dG.GRID_NORMAL_CELL_WIDTH));
+        
+
         // Frame scripts dot
         if(this.model.hasContentfulScripts) {
             ctx.fillStyle = Wick.GUIElement.FRAME_SCRIPT_DOT_COLOR;
             ctx.beginPath();
-            ctx.arc(this.gridCellWidth/2, 0, Wick.GUIElement.FRAME_CONTENT_DOT_RADIUS*1.3, 0, Math.PI);
+            ctx.arc(this.gridCellWidth/2, 0, Wick.GUIElement.FRAME_CONTENT_DOT_RADIUS * _dotScale * 1.3, 0, Math.PI);
             ctx.fill();
         }
 
@@ -122,17 +134,6 @@ Wick.GUIElement.Frame = class extends Wick.GUIElement {
             }
             ctx.lineWidth = Wick.GUIElement.FRAME_CONTENT_DOT_STROKE_WIDTH;
 
-            // dot radius: 22 >> .75; 38 >> 1; 62 >> 1.25
-            var _dcw = this.gridCellWidth;
-            var _dG = Wick.GUIElement;
-            var _dotScale;
-            if (_dcw <= _dG.GRID_SMALL_CELL_WIDTH) {
-                _dotScale = 0.5 + 0.25 * (_dcw / _dG.GRID_SMALL_CELL_WIDTH);
-            } else if (_dcw <= _dG.GRID_NORMAL_CELL_WIDTH) {
-                _dotScale = 0.75 + 0.25 * ((_dcw - _dG.GRID_SMALL_CELL_WIDTH) / (_dG.GRID_NORMAL_CELL_WIDTH - _dG.GRID_SMALL_CELL_WIDTH));
-            } else {
-                _dotScale = 1.0 + 0.25 * ((_dcw - _dG.GRID_NORMAL_CELL_WIDTH) / (_dG.GRID_LARGE_CELL_WIDTH - _dG.GRID_NORMAL_CELL_WIDTH));
-            }
             var r = Wick.GUIElement.FRAME_CONTENT_DOT_RADIUS * _dotScale;
 
             ctx.beginPath();
--- a/engine/src/gui/Frame.js
+++ b/engine/src/gui/Frame.js
@@ -109,7 +109,7 @@ Wick.GUIElement.Frame = class extends Wick.GUIElement {
             ctx.restore();
         }
 
-        if(this.model.tweens.length === 0 && !this.model.sound) {
+        if(this.model.tweens.length === 0 && !this.model.sound && !Wick.GUIElement.HIDE_CONTENT_DOTS) {
             // Frame contentful dot
 
             ctx.fillStyle = Wick.GUIElement.FRAME_CONTENT_DOT_COLOR;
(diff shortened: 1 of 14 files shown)


StickmanRed:
Awesome changes! The new frame size setting works and saves after reloading the project.

Minor UI suggestion: when the frame width goes below a certain threshold, maybe the intermediate gray numbers could be hidden to reduce clutter
image
The size of the playhead handle and contentful frame circle thing don't interpolate as smoothly

https://github.com/user-attachments/assets/b01a56b7-75fc-41e2-82f1-3a2598394cb6

Also the orange frame scripts indicator overflows on extra-small frame sizes, but imo it doesn't matter as much

engine/src/gui/Frame.js
StickmanRed: Maybe make the content dot size dependent on `Wick.GUIElement.GRID_DEFAULT_CELL_WIDTH` (or alias `this.gridCellWidth`), I think this could fix the size issue and possibly make `HIDE_CONTENT_DOTS` able to be removed

And here [Playhead.js](https://github.com/Candlestickers/Candlestick/blob/c1f0aa8423e507922b63a0d4e0fddc0fc46a0c0c/engine/src/gui/Playhead.js#L30-L37)

src/Editor/Modals/SettingsModal/EditorSettings/EditorSettings.jsx
```diff
@@ -126,38 +132,58 @@ class EditorSettings extends Component {
...
+                    h = Math.round(G.GRID_SMALL_CELL_HEIGHT + t * (G.GRID_NORMAL_CELL_HEIGHT - G.GRID_SMALL_CELL_HEIGHT));
+                  } else {
+                    const t = (v - 100) / 50;
+                    w = Math.round(G.GRID_NORMAL_CELL_WIDTH  + t * (G.GRID_LARGE_CELL_WIDTH  - G.GRID_NORMAL_CELL_WIDTH));
+                    h = Math.round(G.GRID_NORMAL_CELL_HEIGHT + t * (G.GRID_LARGE_CELL_HEIGHT - G.GRID_NORMAL_CELL_HEIGHT));
+                  }
+                  G.GRID_DEFAULT_CELL_WIDTH  = w;
+                  G.GRID_DEFAULT_CELL_HEIGHT = h;
+                  // below value 15, hide the content dots
+                  G.HIDE_CONTENT_DOTS = v < 30;
```
StickmanRed: Small error: in `Editor.jsx` this is `G.HIDE_CONTENT_DOTS = v < 15;`

engine/src/gui/Frame.js
StickmanRed:
```javascript
var r = Wick.GUIElement.FRAME_CONTENT_DOT_RADIUS,
    w = this.gridCellWidth,
    factor = 1;
if(w <= Wick.GUIElement.GRID_SMALL_CELL_WIDTH) {
    factor = 0.75;
} else if(w <= Wick.GUIElement.GRID_NORMAL_CELL_WIDTH) {
    factor = 0.75 + 0.25 * (w - Wick.GUIElement.GRID_SMALL_CELL_WIDTH) / (Wick.GUIElement.GRID_NORMAL_CELL_WIDTH - Wick.GUIElement.GRID_SMALL_CELL_WIDTH)
} else {
    factor = 1 + 0.25 * (w - Wick.GUIElement.GRID_NORMAL_CELL_WIDTH) / (Wick.GUIElement.GRID_LARGE_CELL_WIDTH - Wick.GUIElement.GRID_NORMAL_CELL_WIDTH);
}
r *= factor;
```

I think this should do the trick! Maybe we could also make constants of the extra-small width+height and the threshold to hide the content dot. And maybe we could reuse this threshold to hide the gray numbers too

InternetAstronaut:
Hey I got a question, I never really understood the `GUIElement` system in Wick, would it be possible to somehow make it available in a simpler manner for game creators on Wick?

Alani25:
> Hey I got a question, I never really understood the `GUIElement` system in Wick, would it be possible to somehow make it available in a simpler manner for game creators on Wick?

@InternetAstronaut 
GUIElement is more of an internal system for rendering the _timeline editor_ itself (the frames, playhead, layers panel, etc)
The game runs in its own canvas element separately. We can possibly give the user more access to that

Alani25:
@StickmanRed I went for more of an interpolating approach in terms of the sizing, but I did also fix the mismatch issue you brought up. 

https://github.com/user-attachments/assets/5c5bbb70-ddd3-465c-9f57-efbabae823b9

(The suggestion to hide the gray text actually made a huge difference)

InternetAstronaut:
> > Hey I got a question, I never really understood the `GUIElement` system in Wick, would it be possible to somehow make it available in a simpler manner for game creators on Wick?
> 
> @InternetAstronaut GUIElement is more of an internal system for rendering the _timeline editor_ itself (the frames, playhead, layers panel, etc) The game runs in its own canvas element separately. We can possibly give the user more access to that

Huh, well that's weird since all `Wick.Base` objects contain a `View` and a `GUIElement`.

StickmanRed (approved):
Moving frames at a compact frame size is much easier with the `_mouseOverFrameEdge` changes as well! Overall looks good to me

I did notice the onion skin range acts out in the smaller frame sizes, maybe the playhead margin [here](https://github.com/Candlestickers/Candlestick/blob/65f331ec642a1b2ca592bb31afbed3e8b77ed4e3/engine/src/gui/OnionSkinRange.js#L43) could get eased out to 0

StickmanRed:
Yeah, it definitely looks better
image image

```javascript
// In OnionSkinRange.js:
var margin = Wick.GUIElement.PLAYHEAD_MARGIN * 2;
if (this.gridCellWidth < Wick.GUIElement.GRID_SMALL_CELL_WIDTH)
    margin *= (this.gridCellWidth - 8) / (Wick.GUIElement.GRID_SMALL_CELL_WIDTH - 8);
var edgeWidth = this.gridCellWidth - margin;
```

`OnionSkinRange.js` just needs something along these lines, to interpolate `margin` to 0

BaronAWC:
Looks good so far! I have a couple suggestions though:
1. Maybe it's just me, but I think the frame dot should still be shown if there is content in the frame. This is mainly to add some more visual distinction between filled and empty frames.
    a. Something to be careful of, though, is that if a frame is named, the name text and the frame dot will overlap. Not sure what is the smoothest way to fix it. The text needs to show on top for sure, and maybe there's a way to improve the contrast, or we get rid of the dot if there is text.
3. if a frame is too narrow (e.g. <= 16 or 20 pixels wide) and is not currently selected, stretching/shrinking the frame should be disabled, only leaving the ability to select it (duh) and drag it around. the click targets are so small that it's kinda hard to select what you want.
4. The orange code indicator is not currently being resized. It should be set to the same size as the frame bubble.
5. The layers have bad padding. I'm not sure what the best way to deal with it is, but there are a few options:
    a. reduce the margin between layers themselves
    b. don't let the layers/frames get that thin
    c. shrink the buttons and text
    image
6. I think in video editors, you can use a keyboard shortcut to change the size of frames? I forget how it is in DaVinci Resolve, but since normal scroll is for up/down and <kbd>shift</kbd>+scroll is for left/right, we can use either <kbd>alt</kbd>+scroll or <kbd>meta</kbd>+scroll to change the size of frames.

InternetAstronaut:
>I think in video editors, you can use a keyboard shortcut to change the size of frames? I forget how it is in DaVinci Resolve, but since normal scroll is for up/down and shift+scroll is for left/right, we can use either alt+scroll or meta+scroll to change the size of frames.

<kbd>Shift</kbd> + <KBD>,</kbd> or <kbd>.</kbd>. This is simple, and already fits the current shortcuts.

Alani25:
@StickmanRed thanks for pointing out the code, I fixed the onion skinning handles issue thanks to that.
@BaronAWC as per your review, I made the following changes:
- Green dot now shows if the frame has content. 
- When frame size is too small, dragging the frame edges to stretch the frame has now been disabled for all frames except for the active/ selected frames.
- The orange code indicator now resizes the same way as the frames.
- The layers' padding issue was fixed by increasing the frame minimum height.

As per @InternetAstronaut comment in regards to the note on frame extending with the <kbd>shift</kbd> + <kbd>,</kbd> and <kbd>shift</kbd> + <kbd>.</kbd> shortcuts, which I actually realized do exactly this. However they only worked on specifically a _selected frame_, so I made the shortcut also work when no frame is selected, to extend the _active frame_ (frame where the current timeline is at on the last active layer).

BaronAWC:
Just to clarify, when I said "changing the size of frames", I actually meant changing the timeline zoom (my mistake for not specifying). I just opened DaVinci Resolve and can confirm that alt + scroll-up zooms the timeline in, and alt + scroll-down zooms the timeline out. I think it could make sense to add this behavior to the Candlestick timeline too.

SortLock:
*clap clap*

PR #211, opened 2026-09-09
Inspector pivot coordinates and mobile-friendly skew

StickmanRed:
## Description
This PR adds:

- "Pivot X" and "Pivot Y" values to the inspector, both desktop and mobile. These are expressed in relative coordinates to the selection or single clip (before translation, scale, shear, and rotation are applied).

- Pivot snapping to the selection's corners, edge midpoints, and center. It can be disabled by holding `Alt` while dragging the pivot. The pivot snap distance is currently set to the pivot handle's radius.

- The ability to skew by dragging horizontally. This gets overridden when any modifier keys (`Ctrl`, `Shift`, `Alt`) are pressed. Currently set to skew within a vertical distance of 10

https://github.com/user-attachments/assets/63f19903-3774-4c21-95b7-287eedf47882

## Todo
- [ ] Add icons to mobile inspector pivot X/Y inputs

## Testing
Try entering different values for selection pivot X/Y in the inspector. Then move, scale, shear, and rotate the selection using the canvas tool: the values for pivot X/Y should not change.
Drag the pivot point close to the center, one of the corners, or one of the midpoints: at a close enough distance, the pivot should snap to these corners. This can be verified by viewing the pivot's exact coordinates in the inspector.

## Checklist
- [ ] This PR does not make sense to split up into smaller PRs. (I probably should've moved the skewing behavior to a separate PR but oh well)

### DIFF
--- a/engine/src/base/Selection.js
+++ b/engine/src/base/Selection.js
@@ -100,6 +100,7 @@ Wick.Selection = class extends Wick.Base {
             "height",
             "rotation",
             "shear",
+            "relativePivot",
             "opacity",
             "sound",
             "soundVolume",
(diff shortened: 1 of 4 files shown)


BaronAWC (changes requested):
1. When all shortcuts are released while resizing or shearing, dragging perpendicularly to shear gets disabled. I think it'd make sense to still allow shearing, as if the user never pressed anything.
2. If alt is held and the user drags perpendicularly, it should be treated as a shear with the opposite edge staying in place (as if you held meta+alt).
3. Clips are currently the only thing that remember the pivot point, so it doesn't make sense for anything else to have a pivot point parameter. (This can change if we allow objects like paths to not flatten, meaning they'd be able to remember the pivot)

PR #218, opened 2026-09-13
export selected clip to object

plygon1:
added a shortcut where when you press ctrl+shift+e with a clip selected, it exports the clip to a wickobj then either uses the clip's identifier as the name or window.prompts you for the name if the clip doesn't have an identifier. then, that wickobj will be immediately put straight back into your project. this basically lets you add clips to libraries to use across entire projects. this is essentially ctrl+e with extra steps and instead of downloading to your computer, it downloads to the wick project.

this is simply a quality of life feature, and this is my first pull request


InternetAstronaut:
Redundant, #172 also has a similar same premise.

plygon1:
> Redundant, #172 also has a similar same premise.

alright

PR #220, opened 2026-09-15
fix: Clicking "Try it!" welcome button on mobile closes autosave modal

StickmanRed:
## Description
So you press the "Try it!" button, which triggers an `onTouchStart` event. The welcome modal closes and the autosave modal opens. But because the button was a child of the overlay element, you've technically pressed the overlay. And then releasing triggers the overlay element's `click` event, closing the autosave modal before you have a chance to interact with it.

This fix reuses the `useClickEvent` property of buttons from #90 so that the `click` event gets captured by the "Try it!" button, not the outside overlay.

## Testing
1. Use a mobile device or emulate one using Devtools' device toolbar.
2. Reload the Candlestick page.
3. Click the "Try it!" button in the Welcome Modal. The autosave modal should appear after the welcome modal as expected.

## Checklist
- [x] This PR does not make sense to split up into smaller PRs.


BaronAWC (approved):
I've never encountered this problem on my actual phone, but from requesting the mobile version on my laptop, I can get the issue on the main editor and the fix on this branch.

PR #221, opened 2026-09-17
Remove scss import deprecation warnings and fix require('classnames') in ESM code

hobbsythe6th:
Replaced all the `@import` calls in the scss with `@use`. Also fixed how most of the UI code does require('classnames').

This is intended to make the Vite migration easier to read.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "react-modal": "^3.16.3",
     "react-numeric-input": "^2.2.3",
     "react-reflex": "^3.0.6",
-    "react-resize-detector": "^12.3.0",
+    "react-resize-detector": "^9.0.0",
     "react-rnd": "^10.5.2",
     "react-scripts": "^5.0.1",
     "react-select": "^5.9.0",


src/Editor/Editor.jsx
```diff
@@ -23,6 +23,9 @@ import './_editor.scss';
 import './styles/default_theme.css';
 import './styles/default_styles.css';
 
+import version from '../../package.json';
```
StickmanRed: This caused an error in the WelcomeModal, since `import version from '../../package.json` imported the whole object, not just the `version` value; we could change this to
```suggestion
import packageJSON from '../../package.json';
```
and then replace `version` with `packageJSON.version` wherever it's used

src/Editor/Editor.jsx
StickmanRed:
```suggestion
        this.editorVersion = packageJSON.version + '';
```
This is the only usage of `version`

src/Editor/Panels/Canvas/Canvas.jsx
```diff
@@ -63,7 +62,7 @@ class Canvas extends Component {
 
     this.currentAttachedProject = project;
 
-    project.view.canvasBGColor = styles.editorCanvasBorder;
+    project.view.canvasBGColor = '#6A6A6A';
```
StickmanRed: Does Vite disallow importing styles from `./_canvas.scss`? It would be nice to keep the style as a variable, but if it can't, then it's okay. Maybe you could add a comment like `project.view.canvasBGColor = '#6A6A6A'; // styles.editorCanvasBorder`

StickmanRed: Does Vite disallow importing styles from `./_canvas.scss`? It would be nice to keep the style as a variable, but if it can't, then it's okay. Maybe you could add a comment like
```suggestion
    project.view.canvasBGColor = '#6A6A6A'; // styles.editorCanvasBorder
```

StickmanRed (changes requested):
Thanks for all your work on this! Just a couple comments but other than that looks good to me

src/Editor/Editor.jsx
InternetAstronaut:
```suggestion
import { version } from '../../package.json';
```

Problem fixed.

StickmanRed: I was confused why this didn't work at first... it seems like [Vite allows this](https://vite.dev/guide/features#json) but [CRA/Webpack doesn't](https://stackoverflow.com/questions/64993118/error-should-not-import-the-named-export-version-imported-as-version). The named import change is cleaner but has to wait for the main CRA to Vite PR

InternetAstronaut: It's because Old Webpack doesn't fully understand destructuring, it's quite new (ES2016).

hobbsythe6th: I remember that the destructure was there before but it didn't work and also I have NO idea how importing a JSON file works since this is just a helper PR and not standalone. I think destructure works in Vite

StickmanRed (approved):

